PR #14, opened 2024-03-06
Test for adding, removing and modifying value

### LINKED ISSUE #9: Add test cases for diffing changes to values

- [ ] Adding a value
- [ ] Removing a value
- [ ] Modifying an existing value

Those tests should be added to `tests/api-diff/run.t`. We can reuse the base `ref.cmi` as the reference interface. For each test case we should generate a new `.mli` file with a change from the list above, compile it to a `.cmi` with `ocamlc` and then run `api-diff ref.cmi <test-file.cmi>`.

harshey1103:
This PR is wip.

Closes #9


NathanReb:
Yes this is exactly what is expected. From there you should run `dune runtest` and make sure the output corresponds to what we expect. If it does, you can promote the output to the file with `dune promote`. Indeed, the cram test files contains both the command and the expected output.

This is all described in our [contributing guide](https://github.com/NathanReb/ocaml-api-watch/blob/main/CONTRIBUTING.md#writing-tests) and in the [dune documentation](https://dune.readthedocs.io/en/latest/tests.html#cram-tests).

Also before doing this you might want to update your `main` branch and rebase on top of it as we merged a PR updating to OCaml 4.14 and you'll want to build and run the tests with that version.

NathanReb:
You also probably want to change your editor settings to add newlines at the end of files as you'll see your changes removed it.

harshey1103:
While running `dune runtest`, the test for removing a val works fine but for modifying and adding a value the following is generated
```
+  api-watcher: internal error, uncaught exception:
+               Includemod.Error(_)
+               
+  [125]
```
I tried but was not able to solve this. Also how to update Ocaml version to 4.14. I ran `opam switch create ./ 4.12.1 --deps-only -t` this and it created a switch with Ocaml version 4.12.1

shonfeder:
This is an error in the existing program logic, and proves that the tests are working! :tada:

The underlying reason is that the function we are calling to compare interfaces, `Includemod.signatures`, raises an exception when the inclusion check fails: https://github.com/ocaml/ocaml/blob/trunk/typing/includemod.ml#L1212-L1217

I suggest that for now you just `dune promote` those changes and add them to the test case expectations, and we will deal with handling this more gracefully in a followup.

WDYT @NathanReb?

NathanReb:
Yeah, totally agree! The tests are doing their job here.

Once this is merged we'll open an issue to fix the underlying bug they uncovered!

NathanReb (approved):
Looks good! Thank you!

NathanReb:
I just opened #18. @harshey1103 are you interested in fixing the bug?

PR #15, opened 2024-03-07
Update to remove install error

rmeis06:
Hello! I'm an outreachy applicant. 

While following the instructions in CONTRIBUTING.md I got the following error:
```
Switch invariant: ["ocaml-base-compiler" {= "4.12.1"} | "ocaml-system" {= "4.12.1"}]
[ERROR] Could not determine which packages to install for this switch:
  * No agreement on the version of ocaml:
    - (invariant) → ocaml-base-compiler = 4.12.1 → ocaml = 4.12.1
    - api-watch → ocaml >= 4.14.0
    You can temporarily relax the switch invariant with `--update-invariant'
```

Which I resolved by making the change in the PR to my install. Would this change be correct for the install instructions to be working for others as well?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,7 @@ cd ocaml-api-watch
 And setup a local opam switch to install the right ocaml version along with the
 set of base dependencies:
 ```
-opam switch create ./ 4.12.1 --deps-only -t
+opam switch create ./ 4.14.0 --deps-only -t
 ```
 
 You should also install `ocamlformat` and `merlin` for a better dev experience
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,7 @@ cd ocaml-api-watch
 And setup a local opam switch to install the right ocaml version along with the
 set of base dependencies:
 ```
-opam switch create ./ 4.14.0 --deps-only -t
+opam switch create ./ --deps-only -t
 ```
 
 You should also install `ocamlformat` and `merlin` for a better dev experience


CONTRIBUTING.md
```diff
@@ -18,7 +18,7 @@ cd ocaml-api-watch
 And setup a local opam switch to install the right ocaml version along with the
 set of base dependencies:
 ```
-opam switch create ./ 4.12.1 --deps-only -t
+opam switch create ./ 4.14.0 --deps-only -t
```
shonfeder, line 21: I suggest we just remove specification of the compiler version here, since it is already specified in the `dune-project` file at https://github.com/nathanreb/ocaml-api-watch/blob/59eeb9e4c7e09f344cba7df64313aa1b3c8b5768/dune-project#L18 and by no repeating the version here, we avoid having to keep this (untested) instruction up to date :)

```suggestion
opam switch create ./ --deps-only -t
```

shonfeder (commented):
Hello, @rmeis06! Thanks for the contribution and for catching and correcting this error :D  You are absolutely correct with this fix, but I am suggesting a change to prevent this same problem for arising again next time we update the ocaml version :)

CONTRIBUTING.md
rmeis06: Oh that sounds even better! I've updated to match.

shonfeder (approved):

shonfeder:
Awesome work! Thanks again for the fix :)

NathanReb (approved):
Looks good, thanks!

PR #17, opened 2024-03-07
Add test cases for diffing changes to modules

### LINKED ISSUE #10: Add test cases for diffing changes to modules

- [ ] Adding a module
- [ ] Removing a module
- [ ] Modifying an existing module

Those tests should be added to `tests/api-diff/run.t`. This will require either modifying the existing `ref.mli` file created in that test, or creating a new `mod_ref.mli` file, because `ref.mli` doesn't include a module already. For each test case we should generate a new `.mli` file with a change from the list above, compile it to a `.cmi` with `ocamlc` and then run `api-diff ref.cmi <test-file.cmi>`.

Dezzy12:
I have implemented the test cases for Adding, removing and modifying a module.
Fixes #10 

@NathanReb, @shonfeder, @dinakajoy

### DIFF
--- a/tests/api-diff/run.t
+++ b/tests/api-diff/run.t
@@ -28,3 +28,44 @@ there should be no diff:
 
   $ api-diff ref.cmi ref.cmi
   API unchanged!
+
+# Adding a module
+  $ cat > mod_add.mli << EOF
+  > module Hello :
+  >   sig
+  >     type t = unit = ()
+  >     val equal : t -> t -> bool
+  >     val compare : t -> t -> int
+  >     val to_string : t -> string
+  >   end
+  > EOF
+
+  $ ocamlc mod_add.mli
+
+  $ api-diff ref.cmi mod_add.cmi
+  Api changed!
+
+#Removing a module
+  $ cat > mod_remove.mli << EOF
+  > type t = int
+  > 
+  > val f : t -> string
+  > EOF
+
+  $ ocamlc mod_remove.mli
+
+  $ api-diff ref.cmi mod_remove.cmi
+  API unchanged!
+
+# Modifying a module
+  $ cat > mod_modify.mli << EOF
+  > module Hello : sig
+  >   val f : unit -> unit
+  > end
+  > 
+  > EOF
+
+  $ ocamlc mod_modify.mli
+
+  $ api-diff ref.cmi mod_modify.cmi
+  Api changed!


tests/api-diff/run.t
```diff
@@ -28,3 +28,44 @@ there should be no diff:
 
   $ api-diff ref.cmi ref.cmi
   API unchanged!
+
+# Adding a module
+  $ cat > mod_add.mli << EOF
+  > module Hello :
```
NathanReb, line 34: Here you're not simply adding a module compared to `ref.mli`, you're also removing the type and value that is there.

The test case should showcase a single change between the two `.cmi` files.

tests/api-diff/run.t
```diff
@@ -28,3 +28,44 @@ there should be no diff:
...
+  >   sig
+  >     type t = unit = ()
+  >     val equal : t -> t -> bool
+  >     val compare : t -> t -> int
+  >     val to_string : t -> string
```
NathanReb, line 39: We don't need all this here and should keep the test case as simple as possible. Even an empty module signature should work here I guess:
```suggestion
```

tests/api-diff/run.t
```diff
@@ -28,3 +28,44 @@ there should be no diff:
...
+  >   end
+  > EOF
+
+  $ ocamlc mod_add.mli
+
+  $ api-diff ref.cmi mod_add.cmi
+  Api changed!
+
+#Removing a module
+  $ cat > mod_remove.mli << EOF
```
NathanReb, line 49: Here you're just copying the original `ref.mli` file so you're not removing anything.

It's probably best if you define `mod_ref.mli` and `mod_ref.cmi` files that define a couple of modules so you can use them as a base for the three test cases here.

tests/api-diff/run.t
```diff
@@ -28,3 +28,44 @@ there should be no diff:
...
+  $ cat > mod_modify.mli << EOF
+  > module Hello : sig
+  >   val f : unit -> unit
+  > end
+  > 
+  > EOF
+
+  $ ocamlc mod_modify.mli
+
+  $ api-diff ref.cmi mod_modify.cmi
```
NathanReb, line 70: Same here, you need to make it so the only change between the first and second argument is a modification of an existing module.

NathanReb (changes requested):
It looks promising, you got how the cram test work right and properly generate the `.cmi` files and run the tool correctly but the test cases aren't correct at the moment.

I left a few comments to guide you on how to fix them.

Given `ref.mli` does not define any modules you should write a new `mod_ref.mli` that you would use as the reference for the diff in all of your tests, it would probably make them more readable.

NathanReb:
Remember to look at the CI checks to see if your PR builds correctly and is properly formatted.

When writing or editing tests you'll want to run `dune runtest` to see if it all works. You can read the [relevant part of the contributing guide]().

In particular here it should show an error, just as it did in #14. See https://github.com/NathanReb/ocaml-api-watch/pull/14#issuecomment-1982303593 for what to do with it!

harshey1103:
The pull request has been open for an extended period, and @Dezzy12 hasn't been actively contributing to it. Would it be acceptable for me to take over and continue working on this pull request? @Dezzy12 @NathanReb @shonfeder
I would also get a chance to verify my work on https://github.com/NathanReb/ocaml-api-watch/pull/24.

shonfeder:
Let's give a bit of time for @Dezzy12 to reply with whether they still have the time and energy to see this thru. Thanks, @harshey1103!

Dezzy12:
@harshey1103 please go ahead with it. Encountered some issues that I have to sort. So as not to hold down this issue for too long, you can have it if that's okay by @shonfeder

shonfeder:
@harshey1103, could you please comment on #10 so I can assign that issue to you (GitHub doesn't let us assign issues to people who haven't commented on them).

shonfeder:
Closing in favor of #25

PR #19, opened 2024-03-07
Fix Includemod.error

harshey1103:
Added try catch which outputs API changed when there is Inlcudemod.error.
Closes https://github.com/NathanReb/ocaml-api-watch/issues/18

### DIFF
--- a/bin/api_diff.ml
+++ b/bin/api_diff.ml
@@ -2,15 +2,22 @@ let run (`Ref_cmi reference) (`Current_cmi current) =
   let current = Cmi_format.read_cmi current in
   let reference = Cmi_format.read_cmi reference in
   let typing_env = Env.empty in
-  let coercion =
+  let coercion1 =
     try
       Includemod.signatures typing_env ~mark:Mark_both reference.cmi_sign
         current.cmi_sign
     with
     | Includemod.Error _ -> Tcoerce_none
   in
-  match coercion with
-  | Tcoerce_none -> Printf.printf "API unchanged!\n"
+  let coercion2 =
+    try
+      Includemod.signatures typing_env ~mark:Mark_both current.cmi_sign
+        reference.cmi_sign
+    with
+    | Includemod.Error _ -> Tcoerce_none
+  in
+  match (coercion1, coercion2) with
+  | (Tcoerce_none, Tcoerce_none) -> Printf.printf "API unchanged!\n"
   | _ -> Printf.printf "API changed!\n"
 
 let named f = Cmdliner.Term.(app (const f))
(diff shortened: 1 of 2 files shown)


harshey1103:
Hey @NathanReb, one silly doubt if we want to check whether API is changed or not why don't we match both the interfaces line by line instead of using Includemod?
Also, a few observations Includemod works perfectly for the case where the second interface is a subset of the first interface and for the unchanged case. 
Also, I looked up what Includemod does, it checks if the second signature is a superset of the first one. If that is the case we may want to double-check switching the ref and mod cmi files

shonfeder:
@harshey1103 good work digging in to use of `Includemod`!

Here is my understanding of how this check is currently working. Hopefully it helps address your question: The value returned by `Includemod.signatures` is of type [`Typedtree.module_coercion`](https://github.com/ocaml/ocaml/blob/dfb7abefb4df82804703e6609d957f3dcda64d29/typing/typedtree.mli#L499-L517) and this type represents not just *whether or not* the first signature is a sub-type of the second, but what coercion would need to be applied to the first to allow it to count as an instance of the second. If the value is `Tcoerce_none`, then no coercion is need at all, which should mean that the two signatures are the same. Any other value of this type indicates that there is some difference between the signatures, because some coercion would be required. This is why in the current logic, we only report the API as unchanged if we find `Tcoerce_none` and everything else is an indication of a change.

The substantive work on this project will involve more expansive and detailed analysis of the two signatures, so we will have lots of time to expand on our approach here to give more useful feedback!

shonfeder (approved):
Great work here! Looks good to me :)

harshey1103:
Thanks for the input. If there is any issue that needs to be resolved and is not taken by anyone, assign it to me.

NathanReb (commented):
It looks good, thanks!

As @shonfeder pointed out we only used `Includemod` for a first very basic implementation. It's not even clear we'll want to keep using it in the future for a more complete version.

One thing you could do to improve the readability a bit is use an exception case in the pattern matching. You can turn `coercion` into a `unit -> module_coercion` function and then write:
```ocaml
match coercion () with
| Tcoerce_none -> Printf.printf "API unchanged!\n"
| _ -> Printf.printf "API changed!\n"
| exception Includemod.Error _ -> Printf.printf "API changed!\n"
```

It's not strictly required but makes a nice improvement if you're up for it!

harshey1103:
Yeah, I'll push the commit in a sec.

NathanReb (approved):
Awesome, let's merge!

shonfeder:
> Hey @NathanReb, I have a few observations. Includemod works perfectly for the case where the second interface is a subset of the first interface and for the unchanged case. 
> 
> Also, I looked up what Includemod does, it checks if the second signature is a superset of the first one. If that is the case we may want to double-check switching the ref and mod cmi files

Hello! A quick followup that, iiuc, https://github.com/NathanReb/ocaml-api-watch/pull/16#issuecomment-1985366074 shows that you were right in your expected behavior here and it demonstrates the problem you anticipated and that my assumptions about why this would not be a problem were mistaken! :) 

Fortunately, we will be refining the behavior of our check as we go :)

Issue #8, opened 2024-03-06
Add test cases for diffing changes to types


shonfeder:
Having a few simple test cases for different API changes would make for a great way to monitor progress on `api-diff` initial implementation.

Add tests that will check the behavior of `api-diff` when

- [ ] Adding a type
- [ ] Removing a type
- [ ] Modifying an existing type

Those tests should be added to `tests/api-diff/run.t`. We can reuse the base `ref.cmi` as the reference interface. For each test case we should generate a new `.mli` file with a change from the list above, compile it to a `.cmi` with `ocamlc` and then run `api-diff ref.cmi <test-file.cmi>`.

You can read about the testing methodology used for the tests at https://github.com/NathanReb/ocaml-api-watch/blob/main/CONTRIBUTING.md#writing-tests

MungaSoftwiz:
I love the TDD approach we're taking for this project. I would like to be assigned to this task please.

I'm also kindly requesting for recommendations of resources to learn OCaml and their testing methodology. Thank you.

NathanReb:
For OCaml in general I'd recommend the excellent [Real World OCaml](https://dev.realworldocaml.org/). You'll learn plenty about the language and common tools and libraries.

You'll need to learn a bit about dune too so you can take a look at the documentation [here](https://dune.readthedocs.io/en/latest/index.html).

Here we're using dune's cram tests, I explained how they work a bit [here](https://github.com/NathanReb/ocaml-api-watch/blob/main/CONTRIBUTING.md#writing-tests), you can read dune's relevant documentation section [here](https://dune.readthedocs.io/en/latest/tests.html#cram-tests) and you can look at the test we already have in [`tests/api-diff/run.t`](https://github.com/NathanReb/ocaml-api-watch/blob/main/tests/api-diff/run.t) which is the file you'll have to modify to add those test cases.

MungaSoftwiz:
Thank you for the awesome collection @NathanReb . I'll be going through the resources to complement the YouTube tutorials I has started.

MungaSoftwiz:
I was able to look at the project and this is what I deduced from my understanding.
In this context of API diffing we are referring to changes in module interface. So OCaml exposes types through this interface.
adding: introduction of a new type definition. Extending types that module exposes to the user.
removing: removing type definition from module's interface. Reduces set of exposed types to the users of our API.
modifying: modifying an existing type definition.

[...]

PR #24, opened 2024-03-09
Add 2-way check in run function

### LINKED ISSUE #23: The diff comparison is not catching removed types
As observed by @harshey1103 at https://github.com/NathanReb/ocaml-api-watch/pull/19#issuecomment-1985055868, and then demonstrated by @MungaSoftwiz in #16, the way we are using `Includemod.signatures` does not account for the removal of types.

@NathanReb has noted that we will probably end up moving away from `Includemod.signatures`, but in the meantime we should be able to account for this by also calling `Includemod.signatuer` with the arguments `current` and `reference` arguments flipped. This was also suggested by @harshey1103.

harshey1103:
The run function now verifies coercion both of reference to current and of current to reference.
I have locally tested it for adding, removing, and modifying types and it works fine. Hopefully, there won't be any problem in #8.
Closes #23

### DIFF
--- a/bin/api_diff.ml
+++ b/bin/api_diff.ml
@@ -12,11 +12,10 @@ let run (`Ref_cmi reference) (`Current_cmi current) =
     Includemod.signatures typing_env ~mark:Mark_both current.cmi_sign
       reference.cmi_sign
   in
-  try
-    match (coercion1 (), coercion2 ()) with
-    | Tcoerce_none, Tcoerce_none -> Printf.printf "API unchanged!\n"
-    | _, _ -> Printf.printf "API changed!\n"
-  with Includemod.Error _ -> Printf.printf "API changed!\n"
+  match (coercion1 (), coercion2 ()) with
+  | Tcoerce_none, Tcoerce_none -> Printf.printf "API unchanged!\n"
+  | _, _ -> Printf.printf "API changed!\n"
+  | exception Includemod.Error _ -> Printf.printf "API changed!\n"
 
 let named f = Cmdliner.Term.(app (const f))
 
--- a/bin/api_diff.ml
+++ b/bin/api_diff.ml
@@ -4,14 +4,19 @@ let run (`Ref_cmi reference) (`Current_cmi current) =
   let current = Cmi_format.read_cmi current in
   let reference = Cmi_format.read_cmi reference in
   let typing_env = Env.empty in
-  let coercion () =
+  let coercion1 () =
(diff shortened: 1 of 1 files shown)


bin/api_diff.ml
```diff
@@ -4,14 +4,19 @@ let run (`Ref_cmi reference) (`Current_cmi current) =
...
-  | exception Includemod.Error _ -> Printf.printf "API changed!\n"
+  let coercion2 () =
+    Includemod.signatures typing_env ~mark:Mark_both current.cmi_sign
+      reference.cmi_sign
+  in
+  try
+    match (coercion1 (), coercion2 ()) with
+    | Tcoerce_none, Tcoerce_none -> Printf.printf "API unchanged!\n"
+    | _, _ -> Printf.printf "API changed!\n"
+  with Includemod.Error _ -> Printf.printf "API changed!\n"
```
shonfeder, line 19: Let's keep using the case analysis to catch the exception, as we are currently:

```suggestion
  match (coercion1 (), coercion2 ()) with
  | Tcoerce_none, Tcoerce_none -> Printf.printf "API unchanged!\n"
  | _, _ -> Printf.printf "API changed!\n"
  | exception Includemod.Error _ -> Printf.printf "API changed!\n"
```

shonfeder (changes requested):
Great work! Just one request to keep using the approach to exception handling that Nathan previously requested.

bin/api_diff.ml
harshey1103: Yes. I was trying to do it like this 
```
match (coercion1 (), coercion2 ()) with
  | Tcoerce_none, Tcoerce_none -> Printf.printf "API unchanged!\n"
  | _, _ -> Printf.printf "API changed!\n"
  | Includemod.Error _, Includemod.Error _ -> Printf.printf "API changed!\n"
```
and the build was failing. Now it is fine. Thanks. I have pushed the commit.

shonfeder: Ah yea! I can see how that would make sense! As you probably see from this change now, [exception patterns](https://v2.ocaml.org/manual/patterns.html#sss:exception-match) are basically just syntax sugar for wrapping a `match` expression in a `try`, and "exception cases" match against exceptions that arise when evaluating the matched expression, and so are not related to the structure of the matched expression themself.

shonfeder (approved):
Outstanding. Thank you!

NathanReb:
That's great, thanks!

PR #25, opened 2024-03-09
Add tests for adding, removing, and modifying modules

### LINKED ISSUE #10: Add test cases for diffing changes to modules

harshey1103:
Since ref.mli does not contain a module. Created a mod_ref.mli for testing module changes.
Closes #10

### DIFF
--- a/tests/api-diff/run.t
+++ b/tests/api-diff/run.t
@@ -135,10 +135,8 @@ Run api-diff and check the output
 Here we generate a `.mli` file with a module:
 
   $ cat > mod_ref.mli << EOF
-  > type t = int
-  > type unused_type = string
-  > val f : t -> string
   > module M : sig val x : int end
+  > 
   > EOF
 
 We generate the .cmi file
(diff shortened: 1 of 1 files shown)


harshey1103:
The test file seems a bit messy. Is there something we can do to make it easier to understand? Maybe we could add a module to ref.mli or try another solution?

shonfeder:
> The test file seems a bit messy. Is there something we can do to make it easier to understand? Maybe we could add a module to ref.mli or try another solution?

To make *these& test cases simpler, you could remove the types and values and *only* have a module which is added, removed, and changed. If you don't have reason to object, I suggest we do that here.

For the file as a whole, I think adding a module to the main `ref.mli` will only make things more complicated, since all the other cases will also need to account for it.

We could bread these tests into separate files/directories tho. That might help. I suggest we hold off on this until the tests get more complicated however, which they will do shortly :)

shonfeder (approved):

shonfeder:
Looks great. Thanks!

PR #20, opened 2024-03-08
[WIP] Started library implementation 

Siddhi-agg:
As advised by you, I am creating a PR for #12 so that the you can review my code better. It is still unfinished, but I have made the changes you suggested.

PR #26, opened 2024-03-09
Implementing run function logic in library

### LINKED ISSUE #13: Factor out the core logic of `bin/api_diff.ml` into a the library
Requires #12

Thanks to the library component added in #12, we now have a place to put the core logic from `bin/api_diff.ml`. 

The core logic is currently very simple and found in the `run` function

https://github.com/nathanreb/ocaml-api-watch/blob/e0da5f91a20db5fcf20cf37409b298b61f76b19e/bin/api_diff.ml#L1-L11

We want to extract this logic and use it in the dummy function in `lib/abi_diff.ml` defined in #12, so that we'll end up with

```ocaml
let diff_cmi ~reference ~current = (* the core logic from `run` *)
```

This function will take the two paths to the `.cmi` files and return `false` when we get `Tcoerce_none`. I.e., the last lines of `diff_cmi` will adapt the last lines of `run` to be

```ocaml
  (* ... *)
  match coercion with
  | Tcoerce_none -> false
  | _ -> true
```

and `run` itself will end up

```ocaml
let run (`Ref_cmi reference) (`Current_cmi current) =
  if Abi_diff.diff_cmi ~reference ~current then
    Printf.printf "API changed!\n"
  else
    Printf.printf "API unchanged!\n"
```

Siddhi-agg:
Aims to close #13.

### DIFF
--- a/bin/api_diff.ml
+++ b/bin/api_diff.ml
@@ -1,5 +1,10 @@
 let run (`Ref_cmi reference) (`Current_cmi current) =
-  let coercion = Api_watch_diff.diff_interface ~reference ~current in
+  let current = Cmi_format.read_cmi current in
+  let reference = Cmi_format.read_cmi reference in
+  let coercion =
+    Api_watch_diff.diff_interface ~reference:reference.cmi_sign
+      ~current:current.cmi_sign
+  in
   match coercion with
   | false -> Printf.printf "API unchanged!\n"
   | true -> Printf.printf "API changed!\n"
--- a/bin/api_diff.ml
+++ b/bin/api_diff.ml
@@ -1,13 +1,12 @@
 let run (`Ref_cmi reference) (`Current_cmi current) =
   let current = Cmi_format.read_cmi current in
   let reference = Cmi_format.read_cmi reference in
-  let coercion =
+  let differs =
     Api_watch_diff.diff_interface ~reference:reference.cmi_sign
       ~current:current.cmi_sign
   in
-  match coercion with
-  | false -> Printf.printf "API unchanged!\n"
-  | true -> Printf.printf "API changed!\n"
+  if differs then Printf.printf "API changed!\n"
+  else Printf.printf "API unchanged!\n"
 
 let named f = Cmdliner.Term.(app (const f))
 
--- a/bin/api_diff.ml
+++ b/bin/api_diff.ml
@@ -1,5 +1,3 @@
-let _ = Api_watch_diff.diff_interface ~reference:"reference" ~current:"current"
-
 let run (`Ref_cmi reference) (`Current_cmi current) =
   let coercion = Api_watch_diff.diff_interface ~reference:reference ~current:current
   in match (coercion) with
--- a/bin/api_diff.ml
+++ b/bin/api_diff.ml
@@ -1,6 +1,6 @@
 let run (`Ref_cmi reference) (`Current_cmi current) =
-  let coercion = Api_watch_diff.diff_interface ~reference:reference ~current:current
-  in match (coercion) with
+  let coercion = Api_watch_diff.diff_interface ~reference ~current in
+  match coercion with
   | false -> Printf.printf "API unchanged!\n"
   | true -> Printf.printf "API changed!\n"
 
--- a/bin/api_diff.ml
+++ b/bin/api_diff.ml
@@ -1,18 +1,10 @@
 let _ = Api_watch_diff.diff_interface ~reference:"reference" ~current:"current"
 
 let run (`Ref_cmi reference) (`Current_cmi current) =
-  let current = Cmi_format.read_cmi current in
-  let reference = Cmi_format.read_cmi reference in
-  let typing_env = Env.empty in
-  let coercion1 =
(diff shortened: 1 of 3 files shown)


bin/api_diff.ml
```diff
@@ -4,18 +4,15 @@ let run (`Ref_cmi reference) (`Current_cmi current) =
...
-      reference.cmi_sign
-  in
-  match (coercion1 (), coercion2 ()) with
-  | Tcoerce_none, Tcoerce_none -> Printf.printf "API unchanged!\n"
-  | _, _ -> Printf.printf "API changed!\n"
-  | exception Includemod.Error _ -> Printf.printf "API changed!\n"
+  match (coercion1, coercion2) with
+  | Tcoerce_none, Tcoerce_none -> false
+  | _, _ -> true
+  | exception Includemod.Error _ -> true
```
shonfeder, line 15: The goal is to move all of this logic into the `diff_interface` function, then introduce a conditional here that will check the boolean result of that function to decide which output to print.

Siddhi-agg: @shonfeder I have implemented the changes you requested in the PR. Please review them.

lib/api_watch_diff.ml
```diff
@@ -1 +1,16 @@
-let diff_interface ~reference:_ ~current:_ = false
+let diff_interface ~reference ~current =
+  let current = Cmi_format.read_cmi current in
+  let reference = Cmi_format.read_cmi reference in
```
shonfeder, line 3: I think @NathanReb wanted for the `Cmi_format.read_cmi` calls to remain in the `bin/api_diff.ml`. So this function should end up taking the parsed interfaces, rather the strings, and these two lines should be moves back into the `run` function. I didn't explain this well in the issue, sorry!

The motive here is to keep all the I/O operations in the executable, and have the library just handle the core logic.

shonfeder (changes requested):

lib/api_watch_diff.ml
Siddhi-agg: No issues @shonfeder! I have moved the parsing of reference and current files back to ```bin/api_diff.ml```. I hope this is fine now.

bin/api_diff.ml
```diff
@@ -1,21 +1,13 @@
...
   let reference = Cmi_format.read_cmi reference in
-  let typing_env = Env.empty in
-  let coercion1 () =
-    Includemod.signatures typing_env ~mark:Mark_both reference.cmi_sign
-      current.cmi_sign
-  in
-  let coercion2 () =
-    Includemod.signatures typing_env ~mark:Mark_both current.cmi_sign
-      reference.cmi_sign
+  let coercion =
```
NathanReb, line 4: `coercion` is a bit of cryptic term here and we were using it before because it helped understand how we used `Includemod`. Now that the implementation has been moved to the library, the fact that we use `Includemod` becomes an implementation detail we don't have to know about here anymore.

I suggest using a more meaningful word here such as `differs`, as in "does current differs from reference?".

bin/api_diff.ml
```diff
@@ -1,21 +1,13 @@
...
+    Api_watch_diff.diff_interface ~reference:reference.cmi_sign
+      ~current:current.cmi_sign
   in
-  match (coercion1 (), coercion2 ()) with
-  | Tcoerce_none, Tcoerce_none -> Printf.printf "API unchanged!\n"
-  | _, _ -> Printf.printf "API changed!\n"
-  | exception Includemod.Error _ -> Printf.printf "API changed!\n"
+  match coercion with
+  | false -> Printf.printf "API unchanged!\n"
+  | true -> Printf.printf "API changed!\n"
```
NathanReb, line 10: Since you are matching over a boolean here, note that you could rewrite this using an if-then-else:
```suggestion
  if differs then
    Printf.printf "API changed!\n"
  else
    Printf.printf "API unchanged!\n"
```

That's up to you, I don't personally mind the pattern-matching here!

NathanReb (commented):
Could you please add an `.mli` file for `api_watch_diff`?

Without it, ocaml will make everything in the library public. That's not what we would want and instead want to keep control over what we expose.

It should probably have been done as part of #20 but since we're, we might as well fix it!

For now it's pretty easy as there's only one function so we just want to expose this one.

Siddhi-agg:
> Could you please add an `.mli` file for `api_watch_diff`?

Sure @NathanReb! I will start working on it right away.

bin/api_diff.ml
Siddhi-agg: > Since you are matching over a boolean here, note that you could rewrite this using an if-then-else:
@NathanReb Yeah,I was thinking about doing that but since the project was already using pattern matching for printing, I didn't want to make unnecessary changes. I will do it now.

bin/api_diff.ml
Siddhi-agg: > I suggest using a more meaningful word here such as `differs`, as in "does current differs from reference?".

Sorry, since the logic has been encapsulated in the library, I should have used a more meaningful name. I will rectify it asap. Thanks for pointing it out @NathanReb!

shonfeder (approved):
Looks great to me. Thanks!

PR #30, opened 2024-03-14
ppx_expect tests setup

### LINKED ISSUE #27: Add expect tests setup for the api-watch.diff library
At the moment we have a test suite for the `api-diff` tool using cram tests.

Since we want to also provide libraries within this package and we recently extracted the core logic for diffing OCaml interfaces into `api-watch.diff`, we will need to also have an extensive test suite for the library itself.

Those tests should live in a new `tests/api-watch-diff` folder.

## ppx_expect inline tests

We'd like to use `ppx_expect` tests.

`ppx_expect` tests are very similar to cram tests but with OCaml code instead of command line. You can read a bit about our cram tests [here](https://github.com/NathanReb/ocaml-api-watch/blob/main/CONTRIBUTING.md#writing-tests).

It allows you to write OCaml code that prints to the standard output and to ensure that what's printed corresponds to our expectations.

The content of an expect test file will look like a series of such entries:
```ocaml
let%expect_test "test_addition" =
  let result = 1 + 1 in
  Format.printf "%i" result;
  [%expect {|2|}]
```

The `let%expect_test "name" = body` part is used to declare an individual expect test. The expect test `body` should contain the actual test logic and a series of statements printing the relevant values to stdout, in the above it's the `Format.printf ...` statement, followed by a `[%expect s]` where `s` is a string, corresponding to the expected output of the print above.

When running `dune runtest`, if the test fails, you'll get a diff showing that the actual output differed from the expected one. This change can then be accepted and promoted to the test file with `dune promote` if the new output is correct.

To learn how to set them up you can look at [dune's documentation](https://dune.readthedocs.io/en/latest/tests.html#inline-expectation-tests) and [`ppx_expect`'s documentation](https://github.com/janestreet/ppx_expect?tab=readme-ov-file#introduction).

We'll start by simply providing the test setup for this issue and will leave writing actually meaningful tests for later.
To ensure that everything works you can add a single test that `Api_watch_diff.diff_interface ~reference:[] ~current;[]` is `false`.

Siddhi-agg:
This PR will resolve #27. I have provided the initial setup for ```ppx_expect``` inline tests. Please review.


Siddhi-agg:
@NathanReb I still have to add the ppx_expect dependencies to the project. I am currently trying to understand how to do that. Once I do, I will add them. If you can share some resources for that, that would be really helpful.

dune-project
```diff
@@ -16,5 +16,6 @@
  (synopsis "Libraries and tools to keep watch on your OCaml lib's API changes")
  (depends
   (ocaml (and (>= 4.14.0) (< 5.0.0)))
+  (ppx_expect (and (>= v0.14.2) :with-test))
```
NathanReb, line 19: Is there any specific reason why you added the `v0.14.2` lower bound?

tests/api-watch-diff/dune
```diff
@@ -0,0 +1,6 @@
+(library
+ (name ppx_expect_test)
+ (inline_tests)
+ (preprocess
+  (pps ppx_expect))
+ (libraries ppx_expect.evaluator api_watch_diff))
```
NathanReb, line 6: I don't think you need `ppx_expect.evaluator` here it's meant for the test runner which dune is generating by itself, not for the test file.

tests/api-watch-diff/dune
```diff
@@ -0,0 +1,6 @@
+(library
+ (name ppx_expect_test)
```
NathanReb, line 2: Since we only have one module here, it's best to have this match the name of the test file.

NathanReb (commented):
This looks great, thanks!

I have a few minor comments but once they're fixed this should be good to go.

Could you also rename the test file from `ppx_expect.ml` to `test.ml`. The former can be confusing to newcomers that might think we're defining ppx_expect locally.

dune-project
Siddhi-agg: No, there isn't a reason for that. Would you prefer it without bounds?

tests/api-watch-diff/dune
Siddhi-agg: Oh okay! Thanks for pointing it out. I will remove it.

tests/api-watch-diff/dune
Siddhi-agg: Just want to confirm this:
```
(library
 (name test)
```
This should be the new name right? (After I rename test file from ```ppx_expect.ml``` to ```test.ml```)

dune-project
NathanReb: I took a quick look at the changelog of ppx_expect and it's not entirely clear to me. Let's write it without bound for now and see what the CI's lower bounds build says.

tests/api-watch-diff/dune
NathanReb: yes exactly!

dune-project
NathanReb: It seems the lower bound build is broken anyway. Let's stick to the constraint less dep on ppx_expect for now. We'll update it later if needed!

NathanReb (approved):
Perfect, let's merge!

PR #36, opened 2024-03-16
Port tests for changes to types to expect tests

### LINKED ISSUE #31: Port cram test cases to expect tests
Cram tests for the `api-diff` tool and expect tests for the `api-watch.diff` library have different scopes. The expect tests should cover the core logic, i.e. properly detecting and representing API changes while the cram tests should cover the I/O parts and more concrete usecases.

Both being in a very early stage there are not much difference in the set of features the tool covers on top of the library so most of the test cases in our cram test suite actually test the core logic rather than the tool itself, except for the fact that it does properly parse `.cmi` files.

It would therefore make sense to port those test cases as expect tests. We'll keep the duplicate cram for now and let the two test suites diverge as the tool evolves towards more than a mere wrapper to the library main function.

## How to port from cram to expect tests

Cram and expect tests look very similar if not for the fact that the latter use OCaml and not bash. The expect tests won't rely on external files anymore but instead manipulate OCaml values.

The main task here is to understand how to represent the interfaces described in the cram tests `.mli` files in term of OCaml's `Types.signature` and pass them to `Api_watch_diff.diff_interface` instead of generating `.cmi` files and passing them to `api-diff`.

You can look at the `Types` module from the compiler [here](https://github.com/ocaml/ocaml/blob/trunk/typing/types.mli). In particular you'll want to start from [`Types.signature`](https://github.com/ocaml/ocaml/blob/trunk/typing/types.mli#L623) which is the type expected by `Api_watch_diff.diff_interface`. From there you can follow the other type definitions to figure out how to assemble the right `signature`.
If you have trouble understanding what those types correspond to, you can take a look at the [`Parsetree` module](https://github.com/ocaml/ocaml/blob/trunk/parsing/parsetree.mli#L883) from compiler libs. It covers more and slightly different things than `Types` does but a subset of it is very similar an `Parsetree` is better documented so it's a good place to start.

One invocation of `api-diff` in `run.t` should roughly correspond to one `let%expect_test` in `test.ml`.

Note that you can share code in between tests in the expect test file by declaring functions and constants outside of `let%expect_test` entries. It should make the test more concise than in the cram file as it is easier to factor things in OCaml than it is with text in the cram tests.

Building `Types.signature` values for tests input will likely be more verbose than the simple text version in the cram test but the goal is also to identify what can be simplified and to extract helpers into a separate test library.

## Individual issues

This issue centralizes the description of the work but we've cut it into separate smaller issues:
- [ ] #32 
- [x] #33
- [ ] #34
- [ ] #35 

We'll eventually add more tests here to cover the full range of signature items but this makes up for a good starting set.

harshey1103:
Expect tests for adding, removing and modifying types.
Closes #31

### DIFF
--- a/tests/api-watch-diff/test.ml
+++ b/tests/api-watch-diff/test.ml
@@ -2,3 +2,35 @@ let%expect_test "test_diff_interface" =
   let result = Api_watch_diff.diff_interface ~reference:[] ~current:[] in
   Format.printf "%b" result;
   [%expect {|false|}]
+
+open Types
+
+let%expect_test "api-diff ref.cmi ref.cmi" =
+  let signature1 =
+    [
+      Sig_type
+        ( Global "t",
+          {
+            type_params = [];
+            type_arity = 0;
+            type_kind = Type_abstract;
+            type_private = Public;
+            type_manifest =
+              Some
+                (Btype.newgenty
+                   (Tconstr (Path.Pident (Ident.create "int"), [], ref Mnil)));
+            type_variance = [];
+            type_is_newtype = false;
+            type_expansion = None;
+            type_loc = Location.none;
+            type_attributes = [];
+            type_immediate = Unknown;
+            type_unboxed = Default_unboxed;
+          },
+          Trec_not,
+          Exported );
+    ]
+  in
+  let signature2 = signature1 in
+  Api_watch_diff.diff_interface signature1 signature2;
+  [%expect {| API unchanged! |}]
--- a/tests/api-watch-diff/test.ml
+++ b/tests/api-watch-diff/test.ml
@@ -7,8 +7,8 @@ let%expect_test "test_diff_interface" =
 
 let ref_signature =
   (* Signature for ref.mli:
-     type t = int
-     type unused_type = string *)
+     > type t = int
+     > type unused_type = string *)
   [
     Sig_type
       ( Ident.create_persistent "t",
@@ -85,6 +85,9 @@ let ref_signature =
 
 let%expect_test "Testing API watch on the same signature" =
   let curr_signature =
+   (* Signature for curr.mli:
+      > type t = int
+      > type unused_type = string *)
     [
       Sig_type
         ( Ident.create_persistent "t",
@@ -145,6 +148,10 @@ let%expect_test "Testing API watch on the same signature" =
 
 let%expect_test "Testing API watch on adding a type" =
   let curr_signature =
+   (* Signature for curr.mli:
+      > type t = int
+      > type unused_type = string 
+      > type added_type = float *)
     [
       Sig_type
         ( Ident.create_persistent "t",
@@ -229,6 +236,8 @@ let%expect_test "Testing API watch on adding a type" =
 
 let%expect_test "Testing API watch on removing a type" =
   let curr_signature =
+   (* Signature for curr.mli:
+      > type t = int *)
     [
       Sig_type
         ( Ident.create_persistent "t",
@@ -265,6 +274,9 @@ let%expect_test "Testing API watch on removing a type" =
 
 let%expect_test "Testing API watch on modifying a type" =
   let curr_signature =
+   (* Signature for curr.mli:
+      > type t = float
+      > type unused_type = string *)
     [
       Sig_type
         ( Ident.create_persistent "t",
--- a/tests/api-watch-diff/test.ml
+++ b/tests/api-watch-diff/test.ml
@@ -1,5 +1,4 @@
 open Types
-open Predef
 
 let%expect_test "test_diff_interface" =
   let result = Api_watch_diff.diff_interface ~reference:[] ~current:[] in
@@ -217,7 +216,7 @@ let added_type_sig =
         type_arity = 0;
         type_kind = Type_abstract;
         type_private = Public;
-        type_manifest = Some type_float;
+        type_manifest = Some Predef.type_float;
         type_variance = [];
         type_separability = [];
         type_is_newtype = false;
@@ -271,7 +270,7 @@ let mod_t_sig =
         type_arity = 0;
         type_kind = Type_abstract;
         type_private = Public;
-        type_manifest = Some type_float;
+        type_manifest = Some Predef.type_float;
         type_variance = [];
         type_separability = [];
         type_is_newtype = false;
--- a/tests/api-watch-diff/test.ml
+++ b/tests/api-watch-diff/test.ml
@@ -85,9 +85,9 @@ let ref_signature =
 
 let%expect_test "Testing API watch on the same signature" =
   let curr_signature =
-   (* Signature for curr.mli:
-      > type t = int
-      > type unused_type = string *)
+    (* Signature for curr.mli:
+       > type t = int
+       > type unused_type = string *)
     [
       Sig_type
         ( Ident.create_persistent "t",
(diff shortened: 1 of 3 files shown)


harshey1103:
While running tests or building the following change is made in api-watch.opam file:
```
-  "ppx_expect" {>= "v0.14.2" & with-test}
+  "ppx_expect" {with-test}
```
Also to check whether I am moving in the right direction, I created only one signature for the following .mli file 
```
type t = int 
```
I am not able to make Ident.t identifier for the [signature](https://github.com/ocaml/ocaml/blob/18c4510e6c8693c1feffb4871f429987a1d714ff/typing/types.mli#L625). I checked out [Ident.ml](https://github.com/ocaml/ocaml/blob/18c4510e6c8693c1feffb4871f429987a1d714ff/typing/ident.ml#L22) but getting this error
```
File "tests/api-watch-diff/test.ml", line 14, characters 10-16:
14 |         ( Global "t",
               ^^^^^^
Error: Unbound constructor Global
```

harshey1103:
@NathanReb tests for adding and removing are working fine, but the test for modifying type doesn't recognize any changes in the signature. Is something wrong with my signature?

Siddhi-agg:
@harshey1103 I noticed that you haven't implemented the value signature in your PR and since that was my job, you can refer to the implementation [here](https://github.com/NathanReb/ocaml-api-watch/pull/38/files#diff-e629cb7eb02891f07b57494e4d7ad296e9545ce6b763e4178cae4165e134bd21R69).

tests/api-watch-diff/test.ml
```diff
@@ -1,4 +1,428 @@
...
+      ( Ident.create_persistent "t",
+        {
+          type_params = [];
+          type_arity = 0;
+          type_kind = Type_abstract;
+          type_private = Public;
+          type_manifest =
+            Some
+              (Transient_expr.type_expr
+                 (Transient_expr.create (Tvar (Some "int")) ~level:0 ~scope:0
```
shonfeder, line 24: I suspect one problem here (perhaps the main problem) is that you are creating this as a `TVar` which I think is a type variable like `'a`: 

https://github.com/ocaml/ocaml/blob/303f6e2eb13d3b45341a9ba3e1e971f8df7ca685/typing/types.mli#L65-L67

I think what we actually want is a `Tconstr`:

https://github.com/ocaml/ocaml/blob/303f6e2eb13d3b45341a9ba3e1e971f8df7ca685/typing/types.mli#L79-L81

I think you can see an example of how to construct the integer and float types here

https://github.com/ocaml/ocaml/blob/303f6e2eb13d3b45341a9ba3e1e971f8df7ca685/typing/predef.ml#L67-L70

I wonder if we could/should try to use the `Predef` module in these tests... WDYT, @NathanReb?

NathanReb: I'm not extremely familiar with these parts of the compiler yet. I think a good way to better understand this is to figure out a way to pretty print the content of `.cmi` files.

The compiler as several IR (intermediate representation) and most of them can be pretty printed with the `-d<IR>` flags. E.g. you can get the compiler to pretty print the parsetree for a given file with `-dparsetree`. I couldn't find one that would print the right thing but that's something to look into.

If it doesn't exist we can also simply write our own using `[@@deriving show]` or `[@@deriving pp]`.

Anyway, this is a task for later, I'll look deeper into the `Predef` module and how `.cmi` files are compiled and report back!

tests/api-watch-diff/test.ml
```diff
@@ -1,4 +1,400 @@
...
+        {
+          val_type =
+            Transient_expr.type_expr
+              (Transient_expr.create
+                 (Tarrow
+                    ( Nolabel,
+                      Transient_expr.type_expr
+                        (Transient_expr.create
+                           (Tconstr
+                              ( Path.Pident (Ident.create_predef "t"),
```
NathanReb, line 66: I think predef is reserved for ocaml base types, e.g. types predefined by the compiler, such as `int`, `string`, `nativeint` or `bool`

tests/api-watch-diff/test.ml
```diff
@@ -1,4 +1,400 @@
+open Types
+open Predef
```
shonfeder, line 2: Let's use qualified access to `Predef` rather than opening it, so we can see exactly where we are pulling these values from. So that means removing this and anytime we need a value from it we can do, e.g., `Predef.type_int`.

tests/api-watch-diff/test.ml
```diff
@@ -1,4 +1,400 @@
...
+                        commu_ok ))
+                   ~level:0 ~scope:0 ~id:0);
+            val_kind = Val_reg;
+            val_loc = Location.none;
+            val_attributes = [];
+            val_uid = Uid.internal_not_actually_unique;
+          },
+          Exported );
+    ]
+  in
```
shonfeder, line 314: Here should be able to avoid the repetition by removing the type we want with a filter, e.g. with something like:

 ```suggestion
  let current =
    (* Signature for current.mli:
       > type t = int
       > val f : t-> string *)
    let not_unused_type = function
      | Sig_type (id, _) -> Ident.name id <> "unused_type"
      | _ -> false
    in
    List.filter not_unused_type reference
  in
```

File: tests/api-watch-diff/test.ml [hunk 5]
```diff
@@ -1,4 +1,400 @@
...
+                        type_string,
+                        commu_ok ))
+                   ~level:0 ~scope:0 ~id:0);
+            val_kind = Val_reg;
+            val_loc = Location.none;
+            val_attributes = [];
+            val_uid = Uid.internal_not_actually_unique;
+          },
+          Exported );
+    ]
```
shonfeder, line 393: Finally, in this case we avoid the repetition by mapping over the `reference` to replace the type we want to change. E.g., with something like:

```ocaml
let altered_type = (* definition of new type here *) in
List.map (function | Sig_type (id, _) when Ident.name id = "t" -> altered_type | x -> x) reference
```

shonfeder (changes requested):
Great work on getting everything working! I have some suggestions to help reduce the amount of repetition and boilerplate needed.

I think they make the tests more readable. But an alternative approach that could also work is

- Save each constructed signature element of the `reference` as a named value
- Create the reference as a list of these values, e.g., `[t_type, unused_type; f_val]`
- Reuse these named values when constructing the `current` interfaces

I have a preference for the approach I've suggested in the comments, but either approach would improve readability and maintainability IMO.

shonfeder:
Hello again!

tests/api-watch-diff/test.ml
```diff
@@ -1,4 +1,400 @@
...
+  in
+
+  let result =
+    Api_watch_diff.diff_interface ~reference:ref_signature
+      ~current:curr_signature
```
shonfeder, line 157:
```suggestion
    Api_watch_diff.diff_interface ~reference:ref_signature
      ~current:ref_signature
```

tests/api-watch-diff/test.ml
harshey1103: Done

tests/api-watch-diff/test.ml
```diff
@@ -125,6 +125,14 @@ let val_g_sig =
 
 let ref_signature = [ t_sig; unused_type_sig; val_f_sig ]
 
+let%expect_test "same_signature_test" =
```
NathanReb, line 128: You can write any string here so no need to use `_` you can simply use spaces instead.

Also adding at the `test` is a bit redundant given we are in a test file, inside a `let%expect_test` so you can remove those suffixes as well.

NathanReb (commented):
I only have a single comment but besides that, it looks great! Thanks!

tests/api-watch-diff/test.ml
harshey1103: Done. I did it for all the tests.

NathanReb (approved):
This looks great! Thanks @harshey1103

PR #37, opened 2024-03-16
Port tests for changes to modules to expect tests

### LINKED ISSUE #34: Port tests for changes to modules to expect tests

This should cover tests from [run.t line 135](https://github.com/NathanReb/ocaml-api-watch/blob/main/tests/api-diff/run.t#L135) to [run.t line 195](https://github.com/NathanReb/ocaml-api-watch/blob/main/tests/api-diff/run.t#L195).

Refer to the main issue for details on how to proceed.

MungaSoftwiz:
The PR is a WIP closes #34 

I have made some progress with the expected tests for modules but I keep encountering the error below. In the val_type a `type_expr` is needed but I can seem to find the signature of `type_expr` in the `ocaml/typing/types.mli`. The one present is a declaration on the line https://github.com/ocaml/ocaml/blob/18c4510e6c8693c1feffb4871f429987a1d714ff/typing/types.mli#L58.  So I had to put a `type_desc` as a placeholder. At the moment I'm trying to locate the usage of `type_expr` in the compiler in order to replicate it.

Error:
```
...:~/ocaml-api-watch-fork/tests/api-watch-diff$ dune runtest
Entering directory '/home/munga/ocaml-api-watch-fork'
File "tests/api-watch-diff/test.ml", line 16, characters 17-30:
16 |     { val_type = int_type_expr;
                      ^^^^^^^^^^^^^
Error: This expression has type type_desc
       but an expression was expected of type type_expr
```

### DIFF
--- a/tests/api-watch-diff/test.ml
+++ b/tests/api-watch-diff/test.ml
@@ -7,30 +7,41 @@ let%expect_test "test_diff_interface" =
 open Types
 
 let%expect_test "Initial module signature test" =
-  let module_identifier = Ident.create_persistent "M" in
-  let value_identifier = Ident.create_persistent "int" in
-  let int_type_expr =
-    Tconstr (Path.Pident (value_identifier), [], ref Mnil)
-  in
-  let value_description :value_description =
-    { val_type = int_type_expr;
-      val_kind = Val_reg
-      val_attributes = [];
-      val_loc = Location.none;
-      val_uid = Uid.create ()
-    }
-  in
-  let module_type =
-    Mty_signature [Sig_value (value_identifier, value_description, Public)]
-  in
   let mod_ref_signature : signature =
-    [Sig_module (module_identifier,
-                       Mp_present,
-                       module_type,
-                       Nonrecursive,
-                       Public)]
+    [ Sig_module (
+        Ident.create_persistent "M",
+        Mp_present,
+        {
+          md_type = Mty_signature [
+            Sig_value (
+              Ident.create_persistent "x",
+              {
+                val_type =
+                  Transient_expr.type_expr
+                    (Transient_expr.create (
+                      Tconstr (Path.Pident (Ident.create_persistent "int"), [], ref Mnil))
+                        ~level:0 ~scope:0 ~id:0
+                    );
+                val_kind = Val_reg;
+                val_attributes = [];
+                val_loc = Location.none;
+                val_uid = Uid.internal_not_actually_unique;
+              },
+              Exported
+            )
+          ];
+          md_attributes = [];
+          md_loc = Location.none;
+          md_uid = Uid.internal_not_actually_unique;
+        },
+        Trec_not,
+        Exported
+      )
+    ]
   in
-  let diff_result = Api_watch_diff.diff_interface [mod_ref_signature] in
+  let result =
+    Api_watch_diff.diff_interface ~reference:mod_ref_signature ~current:mod_ref_signature
+  in
+  Format.printf "%b" result;
+  [%expect {|false|}]
 
-  Format.printf "%b" diff_result;
-  [%expect {| API unchanged! |}]
--- a/tests/api-watch-diff/test.ml
+++ b/tests/api-watch-diff/test.ml
@@ -3,11 +3,56 @@ let%expect_test "test_diff_interface" =
   Format.printf "%b" result;
   [%expect {|false|}]
 
-(* Test Api_watch_diff.diff_interface with the initial mod_ref_signature *)
 open Types
 
+(* The reference mod_ref_signature *)
+let mod_ref_signature : signature =
+  [
+    Sig_module
+      ( Ident.create_persistent "M",
+        Mp_present,
+        {
+          md_type =
+            Mty_signature
+              [
+                Sig_value
+                  ( Ident.create_persistent "x",
+                    {
+                      val_type =
+                        Transient_expr.type_expr
+                          (Transient_expr.create
+                             (Tconstr
+                                ( Path.Pident (Ident.create_persistent "int"),
+                                  [],
+                                  ref Mnil ))
+                             ~level:0 ~scope:0 ~id:0);
+                      val_kind = Val_reg;
+                      val_attributes = [];
+                      val_loc = Location.none;
+                      val_uid = Uid.internal_not_actually_unique;
+                    },
+                    Exported );
+              ];
+          md_attributes = [];
+          md_loc = Location.none;
+          md_uid = Uid.internal_not_actually_unique;
+        },
+        Trec_not,
+        Exported );
+  ]
+
+(* Test Api_watch_diff.diff_interface with the reference mod_ref_signature *)
 let%expect_test "Initial module signature test" =
-  let mod_ref_signature : signature =
+  let result =
+    Api_watch_diff.diff_interface ~reference:mod_ref_signature
+      ~current:mod_ref_signature
+  in
+  Format.printf "%b" result;
+  [%expect {|false|}]
+
+(* Test Api_watch_diff.diff_interface with the interface mod_add_signature *)
+let%expect_test "Generate a new .mli file with an additional module" =
+  let mod_add_signature : signature =
     [
       Sig_module
         ( Ident.create_persistent "M",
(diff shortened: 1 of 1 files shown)


MungaSoftwiz:
I have managed to figure out how to express types in `type_expr`. This is an expect test for the the `ref_mod.mli` in our cram tests. Let me know if I'm on the right track please. Thank you.

NathanReb:
Yes you are indeed on the right track and I think this is a correct representation of `Module M : sig val x : int end`!

As you can see it is a bit tedious to represent a very simple interface that way, hence the need for helpers there in the future.

MungaSoftwiz:
Awesome! I'm currently winding up on the other tests. Thanks for the feedback.

MungaSoftwiz:
@NathanReb @shonfeder how should we navigate the issue we have here please?
So while working on the `mod_modify_signature`, I realised that an exception is raised with a backtrace showing the flow of execution leading to the error. What led to this is modifying a type in the mdule signature from an `int` to a `float`. Below is the exception output:
My initial thought was that the `api_watch_diff` might not be detecting the change but I'm not sure.

```
File "tests/api-watch-diff/test.ml", line 1, characters 0-0:
diff --git a/_build/default/tests/api-watch-diff/test.ml b/_build/default/tests/api-watch-diff/test.ml.corrected
index 9fb8835..de9e6be 100644
--- a/_build/default/tests/api-watch-diff/test.ml
+++ b/_build/default/tests/api-watch-diff/test.ml.corrected
@@ -177,4 +177,29 @@ let%expect_test "Modify a module in the interface" =
       ~current:mod_modify_signature
   in
   Format.printf "%b" result;
-  [%expect {|true|}]
+  [%expect.unreachable]
+[@@expect.uncaught_exn {|
+  (* CR expect_test_collector: This test expectation appears to contain a backtrace.
+     This is strongly discouraged as backtraces are fragile.
+     Please change this test to not include a backtrace. *)
+
+  (Misc.Fatal_error)
+  Raised at Misc.fatal_errorf.(fun) in file "utils/misc.ml", line 22, characters 14-31
+  Called from Subst.rename_bound_idents.rename_bound_idents in file "typing/subst.ml", line 538, characters 18-33
+  Called from Subst.force_signature_once' in file "typing/subst.ml", line 661, characters 18-50
+  Called from Lazy_backtrack.force in file "utils/lazy_backtrack.ml", line 34, characters 12-15
+  Re-raised at Lazy_backtrack.force in file "utils/lazy_backtrack.ml", line 40, characters 8-15
+  Called from Subst.force_signature_once in file "typing/subst.ml", line 650, characters 18-65
+  Called from Subst.force_signature in file "typing/subst.ml", line 647, characters 32-57
+  Called from Subst.force_modtype in file "typing/subst.ml", line 604, characters 39-59
+  Called from Includemod.modtypes in file "typing/includemod.ml", line 426, characters 15-50
+  Called from Includemod.signature_components in file "typing/includemod.ml", line 758, characters 16-106
+  Called from Includemod.signatures.pair_components in file "typing/includemod.ml", line 653, characters 10-134
+  Called from Includemod.signatures in file "typing/includemod.ml", line 1206, characters 8-111
+  Called from Api_watch_diff.diff_interface.coercion2 in file "lib/api_watch_diff.ml", line 7, characters 4-70
+  Called from Test__Testt.(fun) in file "tests/api-watch-diff/testt.ml", line 176, characters 4-98
+  Called from Expect_test_collector.Make.Instance_io.exec in file "collector/expect_test_collector.ml", line 234, characters 12-19
+
+  Trailing output
+  ---------------
+  >> Fatal error: Ident.rename x |}]
```

Siddhi-agg:
> My initial thought was that the `api_watch_diff` might not be detecting the change but I'm not sure.

@MungaSoftwiz I don't think that is the problem, because when we use ```dune runtest```, the cram tests are also executed and if this was a problem with the ```api_watch_diff``` the cram test of modification would also have failed. But only our expect tests for modification are failing to detect it.

MungaSoftwiz:
Exactly. That was my initial thought but still I admit I might have been wrong. Hopefully we'll figure the issue out.

shonfeder:
Hi :) Here are some notes on how I am debugging. I am writing this out as you may find it useful to see how this kind of problem can be approached. But please know that it is not expected that you'll be able to figure this kind of thing out, let alone have known the right fix. This is pretty thorny stuff, and you are doing great. My take away from this debugging is that we should probably not be manually creating the signatures this way :smile: 

Looking at the traceback the line

```
+  Called from Api_watch_diff.diff_interface.coercion2 in file "lib/api_watch_diff.ml", line 7, characters 4-70
```

tells us that the last place in our code leading tho this exception is this function call on line 7:

https://github.com/mungasoftwiz/ocaml-api-watch/blob/b3741a992e81073cd944b266c790382d2e3c210e/lib/api_watch_diff.ml#L7

```ocaml
    Includemod.signatures typing_env ~mark:Mark_both current reference
```

So this tells us the exception is being thrown from within the compiler libraries we are using. My hypothesis is then that we are probably passing in something malformed. 

I see in the final location of the trace that the error is being thrown by

https://github.com/ocaml/ocaml/blob/bdda743eb66456b659b9f70922a3cba867b57dd2/utils/misc.ml#L20-L24

And that function is supposed to also write the error to stdout. But I think we are missing that output in our expect test because it is getting covered up by the exception output. So I have changed the test to catch the exception to see what is getting written:

```ocaml
  let result =
    try
    Api_watch_diff.diff_interface ~reference:mod_ref_signature
      ~current:mod_modify_signature
    with Misc.Fatal_error -> false
  in
  Format.printf "%b" result;
  [%expect {||}]
```

Which now shows the following in the diff:

```ocaml
--- a/_build/default/tests/api-watch-diff/test.ml
+++ b/_build/default/tests/api-watch-diff/test.ml.corrected
@@ -179,4 +179,6 @@ let%expect_test "Modify a module in the interface" =
     with Misc.Fatal_error -> false
   in
   Format.printf "%b" result;
-  [%expect {||}]
+  [%expect {|
+    >> Fatal error: Ident.rename x
+    false |}]
File "tests/api-watch-diff/test.ml", line 1, characters 0-0:
```

We can find where this is being raised, but this at least tells us that a fatal error is being encountered when the compiler is trying to do some sort of renaming operation on an identifier "x". The exception is being raised by

https://github.com/ocaml/ocaml/blob/bdda743eb66456b659b9f70922a3cba867b57dd2/typing/ident.ml#L55-L61

```ocaml
let rename = function
  | Local { name; stamp = _ }
  | Scoped { name; stamp = _; scope = _ } ->
      incr currentstamp;
      Local { name; stamp = !currentstamp }
  | id ->
      Misc.fatal_errorf "Ident.rename %s" (name id)
```

tbh, I don't have enough context now to explain why this is arising or what is malformed, but it looks clear that this `rename` function only accepts values of type `Ident.t` that are `Local` or `Scoped`, but by using `create_persistent` we are creating a `Global` idenifier:

https://github.com/ocaml/ocaml/blob/bdda743eb66456b659b9f70922a3cba867b57dd2/typing/ident.ml#L46-L47

```ocaml
let create_persistent s =
  Global s
```

It looks like a solution here may be to use `Ident.create_local` instead.

tests/api-watch-diff/test.ml
```diff
@@ -2,3 +2,205 @@ let%expect_test "test_diff_interface" =
...
+    [
+      Sig_module
+        ( Ident.create_persistent "M",
+          Mp_present,
+          {
+            md_type =
+              Mty_signature
+                [
+                  Sig_value
+                    ( Ident.create_persistent "x",
```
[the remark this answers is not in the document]
MungaSoftwiz: Thank you for the very detailed postmorterm @shonfeder. This is a very nice debugging approach and a good way to think about it. Let me use the `create_local` approach with the updated signature for expect tests. Thanks once again!

MungaSoftwiz:
I have changed the `create_persistent` to `create_local` as recommended by @shonfeder. The tests now pass with no issue. I decided to change all identifiers in addition to line 19 and 65 to ensure  consistent naming convention of the `Sig_value` identifiers.

tests/api-watch-diff/test.ml
```diff
@@ -2,3 +2,179 @@ let%expect_test "test_diff_interface" =
...
+                      },
+                      Exported );
+                ];
+            md_attributes = [];
+            md_loc = Location.none;
+            md_uid = Uid.internal_not_actually_unique;
+          },
+          Trec_not,
+          Exported );
+    ]
```
shonfeder, line 119: We can add the new module to the reference just by adding it to the list:

```suggestion
    let new_module =
      Sig_module
        ( Ident.create_persistent "N",
          Mp_present,
          {
            md_type =
              Mty_signature
                [
                  Sig_value
                    ( Ident.create_local "y",
                      {
                        val_type =
                          Transient_expr.type_expr
                            (Transient_expr.create
                               (Tconstr
                                  ( Path.Pident (Ident.create_persistent "float"),
                                    [],
                                    ref Mnil ))
                               ~level:0 ~scope:0 ~id:0);
                        val_kind = Val_reg;
                        val_attributes = [];
                        val_loc = Location.none;
                        val_uid = Uid.internal_not_actually_unique;
                      },
                      Exported );
                ];
            md_attributes = [];
            md_loc = Location.none;
            md_uid = Uid.internal_not_actually_unique;
          },
          Trec_not,
          Exported );
    in
    new_module :: mod_ref_signature
```

tests/api-watch-diff/test.ml
```diff
@@ -2,3 +2,179 @@ let%expect_test "test_diff_interface" =
...
+let%expect_test "Remove a module from the interface" =
+  let mod_remove_signature : signature = [] in
+  let result =
+    Api_watch_diff.diff_interface ~reference:mod_ref_signature
+      ~current:mod_remove_signature
+  in
+  Format.printf "%b" result;
+  [%expect {|true|}]
+
+(* Test Api_watch_diff.diff_interface with the interface mod_modify_signature *)
```
shonfeder, line 138: I think this comment doesn't clarify anything that isn't already apparent in the test declaration. What do you think?

shonfeder (approved):
Great work! See my suggestion to reduce some duplication. Please also take a look at my comments on #36, which you may find instructive and some of which may apply here (probably only the note about the naming for values passed to labels to take advantage of punning). 

Let me know if you have any questions!

tests/api-watch-diff/test.ml
MungaSoftwiz: Thank you for the suggestions. I'll be working on the recommended changes to reduce the duplication and make the tests more readable.

tests/api-watch-diff/test.ml
MungaSoftwiz: It sure does not. It's a habit of documenting code but I admit sometimes it can be quite too much. I will try to tone it down a bit. Thank you for the input @shonfeder .

NathanReb (commented):
It looks good!

Indeed this whole `Ident` thing is a bit hard to deal with and quite confirms that we need the helper library to properly have the compiler assemble signatures rather than trying to build meaningful ones by hand.

I've been looking into it and it seems the ones we want are local and scoped identifiers. Persistent ones seem to be reserved for compilation units (i.e. modules that have their own files).

Here for values `create_local` seems to be the right approach. For pretty much everything else, `create_scoped` is probably the way to go.

tests/api-watch-diff/test.ml
```diff
@@ -2,3 +2,179 @@ let%expect_test "test_diff_interface" =
...
+        Mp_present,
+        {
+          md_type =
+            Mty_signature
+              [
+                Sig_value
+                  ( Ident.create_local "x",
+                    {
+                      val_type =
+                        Transient_expr.type_expr
```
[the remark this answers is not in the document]
NathanReb: See #36, @harshey1103 used this there!

MungaSoftwiz:
@NathanReb @shonfeder 
I used the `@` for concatenation of the two module signatures as `::` resulted in the creation of a `signature_item list`. The resulting error was:
```
File "tests/api-watch-diff/test.ml", line 274, characters 41-61:
274 | let add_module_signature = new_module :: ref_module_signature
                                               ^^^^^^^^^^^^^^^^^^^^
Error: This expression has type signature_item list
       but an expression was expected of type signature_item list list
       Type signature_item is not compatible with type signature_item list
```
I am aware that `@` is because it has a time complexity of `O(n)` so I'll have to look for a way-around to incorporate `::`

tests/api-watch-diff/test.ml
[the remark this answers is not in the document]
MungaSoftwiz: Interesting. Let me go see that and try to refactor the tests.

NathanReb:
> I am aware that @ is slower because it has a time complexity of O(n) so I'll have to look for a way-around to incorporate ::. Our tests have to be fast as possible.

You can simply define a variable for the individual signature items, e.g.:
```ocaml
let module_m =
  Sig_module
    ( Ident.create_persistent "M",
      Mp_present,
      ... )
 
let module_n =
  Sig_module
    (Ident.create_persistent "N",
     ...)
```

And then assemble the signature as a list literal:
```ocaml
let signature = [module_m; module_n]
```

Also, can you mark the comments as resolved once you have adressed them please. That helps keeping track of what's left to do here.

NathanReb:
Note that in ocaml, the following:
```ocaml
[a; b; c]
```
is syntactic sugar for:
```ocaml
a::b::c::[]
```

MungaSoftwiz:
> is syntactic sugar for:
> 
> ```ocaml
> a::b::c::[]
> ```

Thank you for this @NathanReb. When I used the method above I encountered some errors that were the opposite to what I encountered at https://github.com/NathanReb/ocaml-api-watch/pull/37#issuecomment-2015256365, Here are the exact errors:

1. When using the modules as variables like above and combining them to a list:
```
295 let ref_module_m_signature =
  1     Sig_module
  2       ( Ident.create_persistent "M",
  3         Mp_present,
```
```
356 let add_module_signature = [ module_n_signature; ref_module_m_signature ]
...

360 |     Api_watch_diff.diff_interface ~reference:ref_module_m_signature
                                                   ^^^^^^^^^^^^^^^^^^^^^
Error: This expression has type signature_item
       but an expression was expected of type signature_item list
```

So the trick was to specify one signature as a `signature_item` (module_n_signature) and another as a `signature_item list` (ref_module_m_signature) since the cons operator can't concatenate two lists.
So the other ones are `signature_item list` except **module_n**, which is the added one.

NathanReb:
You don't have to make the ref module a list. Indeed it is a signature item so if you want to build a signature made of that module alone, based on my examples above you could simply write:
```ocaml
let ref_module_signature = [module_m]
```

Making it a list makes it harder to reuse it in other signatures without concatenating list. Making if of a different type can also be confusing here since all other elements or `signature_item`s.

You don't have to use `::` directly and list literals are actually much easier to read in my opinion. The cons operator `::` can be thought of as an infix version of `List.append`, `x::l` will simply append the left hand side argument `x` of type `'a` to the right hand side argument `l` of type `'a list`.

The expression `a::b::c::[]` is in fact `a::(b::(c::[]))`. I think I got you confused with the explanation of how list literals are interpreted so don't bother with the cons operator here. in `[a; b; c]` all of `a`, `b` and `c` are all of the same type `'a` and the whole expression `[a; b; c]` is of type `'a list`.

MungaSoftwiz:
Thank you for the recommendations. Let me fix the issue right away.

NathanReb (approved):
Looks good, thanks!

Issue #35, opened 2024-03-15
Add a test helper library for expect tests


NathanReb:
This is one of the sub issues for #31.

Refer to the main issue for details.

NathanReb:
Given how complex and verbose it is to write our own `Types.signature` it seems that a very useful helper function for writing basic test would be one that takes as input a string corresponding to the content of an `.mli` file, parse it, compiles it to a typedtree and returns the `Types.signature` enclosed.

One could then write tests in a similar fashion to:
```ocaml
let%expect_test "..." =
  let ref_interface =
    {|
type t = int

val f : t -> string
|}
  in
  let modified_interface =
    interface ^
    {|https://v2.ocaml.org/api/String.html
val g : t -> bool
|}
  in
  let reference = Test_helpers.compile_interface ref_interface in
  let current = Test_helpers.compile_interface modified_interface in
  let result = Api_watch_diff.diff_interface ~reference ~current in
  Format.printf "%b" result;
  [%expect false]
```

This can be done using the `Parse`, `Env` and `Typemod` modules from compiler libs.

MungaSoftwiz:
I am interested @NathanReb

harshey1103:
Yes. I am interested and almost done with previous issue.

PR #38, opened 2024-03-16
Port tests for changes to values to expect tests 

### LINKED ISSUE #33: Port tests for changes to values to expect tests

This should cover tests from [run.t line 85](https://github.com/NathanReb/ocaml-api-watch/blob/main/tests/api-diff/run.t#L85) to [run.t line 133](https://github.com/NathanReb/ocaml-api-watch/blob/main/tests/api-diff/run.t#L133).

Refer to the main issue for details on how to proceed.

Siddhi-agg:
Aims to close #33. This is my initial work on the issue. I have opened this PR to record my progress and display it for ease in reviewing.

### DIFF
--- a/tests/api-watch-diff/test.ml
+++ b/tests/api-watch-diff/test.ml
@@ -19,8 +19,10 @@ let t_sig =
         type_manifest =
           Some
             (Transient_expr.type_expr
-               (Transient_expr.create (Tvar (Some "int")) ~level:0 ~scope:0
-                  ~id:0));
+               (Transient_expr.create
+                  (Tconstr
+                     (Path.Pident (Ident.create_predef "int"), [], ref Mnil))
+                  ~level:0 ~scope:0 ~id:0));
         type_variance = [];
         type_separability = [];
         type_is_newtype = false;
@@ -48,8 +50,10 @@ let unused_type_sig =
         type_manifest =
           Some
             (Transient_expr.type_expr
-               (Transient_expr.create (Tvar (Some "string")) ~level:0 ~scope:0
-                  ~id:1));
+               (Transient_expr.create
+                  (Tconstr
+                     (Path.Pident (Ident.create_predef "string"), [], ref Mnil))
+                  ~level:0 ~scope:0 ~id:1));
         type_variance = [];
         type_separability = [];
         type_is_newtype = false;
@@ -76,11 +80,17 @@ let val_f_sig =
                (Tarrow
                   ( Nolabel,
                     Transient_expr.type_expr
-                      (Transient_expr.create (Tvar (Some "t")) ~level:0 ~scope:0
-                         ~id:2),
+                      (Transient_expr.create
+                         (Tconstr
+                            (Path.Pident (Ident.create_predef "t"), [], ref Mnil))
+                         ~level:0 ~scope:0 ~id:2),
                     Transient_expr.type_expr
-                      (Transient_expr.create (Tvar (Some "string")) ~level:0
-                         ~scope:0 ~id:3),
+                      (Transient_expr.create
+                         (Tconstr
+                            ( Path.Pident (Ident.create_predef "string"),
+                              [],
+                              ref Mnil ))
+                         ~level:0 ~scope:0 ~id:3),
                     commu_ok ))
                ~level:0 ~scope:0 ~id:4);
         val_kind = Val_reg;
@@ -103,11 +113,15 @@ let val_g_sig =
                (Tarrow
                   ( Nolabel,
                     Transient_expr.type_expr
-                      (Transient_expr.create (Tvar (Some "t")) ~level:0 ~scope:0
-                         ~id:5),
+                      (Transient_expr.create
+                         (Tconstr
+                            (Path.Pident (Ident.create_predef "t"), [], ref Mnil))
+                         ~level:0 ~scope:0 ~id:5),
                     Transient_expr.type_expr
-                      (Transient_expr.create (Tvar (Some "t")) ~level:0 ~scope:0
-                         ~id:6),
+                      (Transient_expr.create
+                         (Tconstr
+                            (Path.Pident (Ident.create_predef "t"), [], ref Mnil))
+                         ~level:0 ~scope:0 ~id:6),
                     commu_ok ))
                ~level:0 ~scope:0 ~id:7);
         val_kind = Val_reg;
@@ -167,11 +181,15 @@ let modified_val_f_sig =
                (Tarrow
                   ( Nolabel,
                     Transient_expr.type_expr
-                      (Transient_expr.create (Tvar (Some "t")) ~level:0 ~scope:0
-                         ~id:8),
+                      (Transient_expr.create
+                         (Tconstr
+                            (Path.Pident (Ident.create_predef "t"), [], ref Mnil))
+                         ~level:0 ~scope:0 ~id:8),
                     Transient_expr.type_expr
-                      (Transient_expr.create (Tvar (Some "t")) ~level:0 ~scope:0
-                         ~id:9),
+                      (Transient_expr.create
+                         (Tconstr
+                            (Path.Pident (Ident.create_predef "t"), [], ref Mnil))
+                         ~level:0 ~scope:0 ~id:9),
                     commu_ok ))
                ~level:0 ~scope:0 ~id:10);
         val_kind = Val_reg;
@@ -194,4 +212,4 @@ let%expect_test "modifying_a_value_test" =
       ~current:modify_value_signature
   in
   Format.printf "%b" result;
-  [%expect {|false|}]
+  [%expect {|true|}]
--- a/tests/api-watch-diff/test.ml
+++ b/tests/api-watch-diff/test.ml
@@ -2,3 +2,101 @@ let%expect_test "test_diff_interface" =
   let result = Api_watch_diff.diff_interface ~reference:[] ~current:[] in
   Format.printf "%b" result;
   [%expect {|false|}]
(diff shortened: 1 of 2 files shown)


Siddhi-agg:
@NathanReb I had a question. When I execute ```dune runtest``` I get the following error : 
```
......................................{
                  desc = Tconstr (Path.Pident "int", [], ref []);
                  level = 0;
                  scope =0;
                  id = 1;
                };
Error: Cannot create values of the private type Types.transient_expr
```
Then, when I change the code to:
```
 let a = 0 in
  let signature1=
  [
   .... 
        type_manifest = Some {
                  desc = Tconstr (Path.Pident "int", [], ref []);
                  level = 0;
                  scope = Some a;
                  id = 1;
                };
``` 
It still gives the same error. Shouldn't changing the scope from 0 to Some a (0) change its type from ```int``` to ```int option``` and thus change the whole expression's type to type_expr?

harshey1103:
Hi @Siddhi-agg, I encountered a similar error to the one you mentioned. Since I am responsible for implementing the type part. You can refer [here](https://github.com/NathanReb/ocaml-api-watch/pull/36/files#diff-e629cb7eb02891f07b57494e4d7ad296e9545ce6b763e4178cae4165e134bd21R18) for the fix.

NathanReb:
Indeed @Siddhi-agg, your error was that you were trying to build a value of a private type.

They are documented [here](https://v2.ocaml.org/manual/privatetypes.html) if you want to take a look but in short, private types are kind of "read-only" types as in you can pattern match over their structure as if they were public but you cannot directly create a value of that type, you have to use API functions to do so just as if they were abstract.

E.g. if you have the following:
```ocaml
type t = private
  | A
  | B of int
  
val a : unit -> t
val b : int -> t
```

That means you can write the following code:
```ocaml
match t with
| A -> ...
| B i -> ...
```

but you cannot write:
```ocaml
let x = B 10
```

You will have to instead use the function provided above:
```ocaml
let x = b 10
```

Siddhi-agg:
> They are documented [here](https://v2.ocaml.org/manual/privatetypes.html) if you want to take a look but in short, private types are kind of "read-only" types as in you can pattern match over their structure as if they were public but you cannot directly create a value of that type, you have to use API functions to do so just as if they were abstract.

Yeah, I read that and used ```create``` function from to get a declare a ```type_expr```. @harshey1103 's implementation helped a lot in that. Thanks :)

I am currently working on why the ```diff_interface``` isn't detecting the modified values. I also noticed that @harshey1103 and @MungaSoftwiz are experiencing the same issue in their respective tests. Would be glad to help them if I solve the issue.

MungaSoftwiz:
> > They are documented [here](https://v2.ocaml.org/manual/privatetypes.html) if you want to take a look but in short, private types are kind of "read-only" types as in you can pattern match over their structure as if they were public but you cannot directly create a value of that type, you have to use API functions to do so just as if they were abstract.
> 
> Yeah, I read that and used `create` function from to get a declare a `type_expr`. @harshey1103 's implementation helped a lot in that. Thanks :)
> 
> I am currently working on why the `diff_interface` isn't detecting the modified values. I also noticed that @harshey1103 and @MungaSoftwiz are experiencing the same issue in their respective tests. Would be glad to help them if I solve the issue.

Yes that's an issue I have experienced too. From my PR I have used the expect tests as a placeholder for now. I've been trying to see how to navigate that issue better too. 
What's the output on your side please?

Siddhi-agg:
> What's the output on your side please?

@MungaSoftwiz On my side, I just get asked to change the expected output of the modify test to ```false```:
```File "tests/api-watch-diff/test.ml", line 1, characters 0-0:
diff --git a/_build/default/tests/api-watch-diff/test.ml b/_build/default/tests/api-watch-diff/test.ml.corrected
index 491825a..f3d351f 100644
--- a/_build/default/tests/api-watch-diff/test.ml
+++ b/_build/default/tests/api-watch-diff/test.ml.corrected
@@ -180,7 +180,7 @@ let%expect_test "modifying_a_value_test" =
       ~current:ref_signature
   in
   Format.printf "%b" result;
-  [%expect {|true|}]
+  [%expect {|false|}]
 
 (* Signature for remove_value.mli:
      > type t = int
```

MungaSoftwiz:
> > What's the output on your side please?
> 
> @MungaSoftwiz On my side, I just get asked to change the expected output of the modify test to `false`:
> 
> ```
> diff --git a/_build/default/tests/api-watch-diff/test.ml b/_build/default/tests/api-watch-diff/test.ml.corrected
> index 491825a..f3d351f 100644
> --- a/_build/default/tests/api-watch-diff/test.ml
> +++ b/_build/default/tests/api-watch-diff/test.ml.corrected
> @@ -180,7 +180,7 @@ let%expect_test "modifying_a_value_test" =
>        ~current:ref_signature
>    in
>    Format.printf "%b" result;
> -  [%expect {|true|}]
> +  [%expect {|false|}]
>  
>  (* Signature for remove_value.mli:
>       > type t = int
> ```

Interesting @Siddhi-agg! It's surprising that a change in types in the modules outputs a more verbose exception.
This bags the question, why is the change of a type in a module causing the particular backtrace compared to values?

tests/api-watch-diff/test.ml
```diff
@@ -2,3 +2,196 @@ let%expect_test "test_diff_interface" =
...
+    ( Ident.create_persistent "t",
+      {
+        type_params = [];
+        type_arity = 0;
+        type_kind = Type_abstract;
+        type_private = Public;
+        type_manifest =
+          Some
+            (Transient_expr.type_expr
+               (Transient_expr.create (Tvar (Some "int")) ~level:0 ~scope:0
```
shonfeder, line 22: I think you are hitting the same problem I explained here https://github.com/NathanReb/ocaml-api-watch/pull/36#discussion_r1529470985 #37 

The gist is that these should be type constructors (`Tconst`) and not type variables (`Tvar`) :)

Siddhi-agg: @shonfeder Thanks for this approach! I think the it worked. Also, we would have to use ```Ident.create_predef``` instead of ```Ident.create_persistent``` when giving the path to ```Tconstr``` because the ```Global``` identifier doesn't let the test detect the changes made to the ```value```.

tests/api-watch-diff/test.ml
```diff
@@ -2,3 +2,214 @@ let%expect_test "test_diff_interface" =
...
+      {
+        val_type =
+          Transient_expr.type_expr
+            (Transient_expr.create
+               (Tarrow
+                  ( Nolabel,
+                    Transient_expr.type_expr
+                      (Transient_expr.create
+                         (Tconstr
+                            (Path.Pident (Ident.create_predef "t"), [], ref Mnil))
```
NathanReb, line 118: I think you shouldn't be using `create_predef` here. `t` is not a predefined OCaml type.

tests/api-watch-diff/test.ml
```diff
@@ -2,3 +2,214 @@ let%expect_test "test_diff_interface" =
...
+        type_params = [];
+        type_arity = 0;
+        type_kind = Type_abstract;
+        type_private = Public;
+        type_manifest =
+          Some
+            (Transient_expr.type_expr
+               (Transient_expr.create
+                  (Tconstr
+                     (Path.Pident (Ident.create_predef "string"), [], ref Mnil))
```
NathanReb, line 55: I think you can directly use `Predef.path_string` here. Same apply to `int` or any predefined ocaml type.

tests/api-watch-diff/test.ml
```diff
@@ -2,3 +2,214 @@ let%expect_test "test_diff_interface" =
...
+  Sig_type
+    ( Ident.create_persistent "t",
+      {
+        type_params = [];
+        type_arity = 0;
+        type_kind = Type_abstract;
+        type_private = Public;
+        type_manifest =
+          Some
+            (Transient_expr.type_expr
```
NathanReb, line 21: I think you can use `Typing.create_expr` directly here instead of `Transient_expr. (type_exr (create ...))`.

NathanReb (commented):
Thanks! This is very useful work you're doing as it helps us dive into the compiler's representation of interfaces.

It also helped us figure out that in its current form it is neither a satisfying nor a scalable way to write tests and we need helpers to build signatures.

I still think it's worth going through with this before #35. I left a few comments for some improvements but besides that it looks good.

Siddhi-agg:
> It also helped us figure out that in its current form it is neither a satisfying nor a scalable way to write tests and we need helpers to build signatures.

@NathanReb I agree. When I was building signatures for tests, even I thought that was a tedious and verbose process to do manually for the various tests this project may have in the future. Having a helper library to automate generating signatures would be nice!

tests/api-watch-diff/test.ml
```diff
@@ -2,3 +2,198 @@ let%expect_test "test_diff_interface" =
...
+        val_uid = Uid.internal_not_actually_unique;
+      },
+      Exported )
+
+(* Signature for file ref.mli:
+      > type t = int
+      > type unused_type = string
+      > val f : t -> string *)
+
+let ref_signature = [ t_sig; unused_type_sig; val_f_sig ]
```
shonfeder, line 126: This is a really nice approach, and it saves having to write a bunch of duplicate code. It makes your tests very easy to read.

shonfeder (approved):
Looks great!

tests/api-watch-diff/test.ml
Siddhi-agg: Thanks @shonfeder :)

PR #39, opened 2024-03-19
Nawaraj upgrade version

nawarajshah:
Upgrade to OCaml 5.1

Issue: https://github.com/NathanReb/ocaml-api-watch/issues/28

I am making this new PR because even though I try, I can not solve it. So here I start from the start.


NathanReb (approved):
This looks good, thanks!

PR #42, opened 2024-05-27
Refactored tests using test_helpers library

Siddhi-agg:
This PR gives us the rewritten ppx-expect-tests. The new tests have been written using the test_helpers library, which has been applied to generate signatures for files automatically for comparison of interfaces. @NathanReb Please review.

### DIFF
--- a/tests/api-watch-diff/test.ml
+++ b/tests/api-watch-diff/test.ml
@@ -1,119 +1,20 @@
-open Types
-
 let%expect_test "test_diff_interface" =
   let result = Api_watch_diff.diff_interface ~reference:[] ~current:[] in
   Format.printf "%b" result;
   [%expect {|false|}]
 
-(* Signature for type t:
-     > type t = int *)
-
-let t_sig =
-  Sig_type
-    ( Ident.create_persistent "t",
-      {
-        type_params = [];
-        type_arity = 0;
-        type_kind = Type_abstract;
-        type_private = Public;
-        type_manifest = Some Predef.type_int;
-        type_variance = [];
-        type_separability = [];
-        type_is_newtype = false;
-        type_expansion_scope = 0;
-        type_loc = Location.none;
-        type_attributes = [];
-        type_immediate = Unknown;
-        type_unboxed_default = false;
-        type_uid = Uid.internal_not_actually_unique;
-      },
-      Trec_not,
-      Exported )
-
-(* Signature for type t:
-     > type unused_type = string *)
-
-let unused_type_sig =
-  Sig_type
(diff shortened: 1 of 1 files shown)


tests/api-watch-diff/test.ml
```diff
@@ -1,119 +1,16 @@
...
-(* Signature for file ref.mli:
+(* Signature for ref.mli:
       > type t = int
       > type unused_type = string
       > val f : t -> string *)
 
-let ref_signature = [ t_sig; unused_type_sig; val_f_sig ]
+let ref_signature =
+  Test_helpers.compile_interface
+    "type t = int\ntype unused_type = string\nval f : t -> string"
```
NathanReb, line 13: since we're going to write OCaml code in those string literals I'd recommend using the `{|` and `|}` string delimiters.

That will save some annoying escaping and allow us to have the string litterals look like actual code:
```suggestion
    {|
    type t = int
    type unused_type = string
    val f : t -> string
    |}
```

NathanReb: Same applies to other occurences of course!

No need to use `\n` either. An actual newline should work!

NathanReb (commented):
Looks good! I added a comment on how to get better looking string literals but other than that it looks good.

When you're done could you also merge the commits together into one. We try to keep meaningful commits. There can be several commits per PR but each usually should correspond to a specific change in the code so that browsing the history and `git blame`-ing gives the user a good idea of what the commit was about.
In particular, try to run `dune build @all @runtest` and `dune build @fmt` before each commit so the code is always valid and properly formatted. That can help a lot when bisecting in search of a bug!

Siddhi-agg:
@NathanReb I have made the suggested changes to the string literals. They do look better with string delimiters!

Also, I have squashed the commits into a single meaningful commit. Not running `dune fmt` before making a commit was a mistake on my part. I will keep note of that in my further contributions. Hope this PR is okay now!

Siddhi-agg:
@NathanReb I ran `dune build @fmt` on my terminal and it updated some of my code. I used 'git add` on the code to incorporate those changes, but now I see the changes were not added. This may be because I use ssh to connect my virtual machine where this project is located and VSCode which is in my local machine.
I really apologise for this inconvenience.

NathanReb:
Running `dune build @fmt` does not reformat, it simply checks whether the file are properly formatted and shows the diff otherwise. You need to run `dune promote` if it showed you a diff to "accept" those changes and promote them to the source files.

You can also run `dune build @fmt --auto-promote` if you want to include the promotion step automatically.

PR #47, opened 2024-06-12
Fixed implementation for diffing values more general than one another

Siddhi-agg:
The goal of this PR is to enhance our implementation to handle diffing of values with more general types than one another, ensuring accurate detection and reporting of modifications.

I have also added simple expect tests for handling the both the cases : 

- when the reference signature is more general than the current signature
- when the current signature is more general than the reference signature

### DIFF
--- a/tests/api-watch-diff/test.ml
+++ b/tests/api-watch-diff/test.ml
@@ -213,3 +213,17 @@ let%expect_test "Modifying a module" =
   in
   Format.printf "%a" pp_diff_list result;
   [%expect {|[Any]|}]
+
+let%expect_test "Reference more general than Current interface" =
+  let reference = Test_helpers.compile_interface {|val x : 'a list|} in
+  let current = Test_helpers.compile_interface {|val x : float list|} in
+  let result = Api_watch_diff.diff_interface ~reference ~current in
+  Format.printf "%a" pp_diff_list result;
+  [%expect {|[Value (x, Modified)]|}]
+
+let%expect_test "Current more general than Reference interface" =
+  let reference = Test_helpers.compile_interface {|val x : int list|} in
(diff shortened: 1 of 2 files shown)


tests/api-watch-diff/test.ml
```diff
@@ -213,3 +213,17 @@ let%expect_test "Modifying a module" =
...
+  let result = Api_watch_diff.diff_interface ~reference ~current in
+  Format.printf "%a" pp_diff_list result;
+  [%expect {|[Value (x, Modified)]|}]
+
+let%expect_test "Current more general than Reference interface" =
+  let reference = Test_helpers.compile_interface {|val x : int list|} in
+  let current = Test_helpers.compile_interface {|val x : 'a list|} in
+  let result = Api_watch_diff.diff_interface ~reference ~current in
+  Format.printf "%a" pp_diff_list result;
+  [%expect {|[Value (x, Modified)]|}]
```
NathanReb, line 229: You could potentially do all of this in the same test here since we're trying to show that the modification is detected both ways. You can have multiple "expect statement" in the same test:
```suggestion
  [%expect {|[Value (x, Modified)]|}];
  let rev_result =
    Api_watch_diff.diff_interface ~reference:current ~current:reference
  in
  Format.printf "%a" pp_dif_list rev_result;
  [%expect {|[Value (x, Modified)]|}]
```

To clarify things we'd need to rename `reference` and `current` to `general` and `specialized` for instance, and change the name of the test to reflect that it's not focused on a single "direction".

NathanReb (commented):
It looks great, thanks!

I made a comment about merging both new tests into one but other than that, we're good to go!

NathanReb (approved):
Wonderful! Let's merge!

PR #48, opened 2024-06-17
Fixed implementation for abstract types

### LINKED ISSUE #44: Improve diffing of values: Abstract types
Our current implementation is able to detect addition, deletion and modification of basic values but it is not complete and will fail to detect some value modifications properly.

We'll fix the implementation incrementally, adding new test cases to our expect tests suite and improving the implementation to support them.

The first and most important one is to make sure that abstract types with the same name are considered equal between the reference and current interfaces when adequate.

Consider the following interface:
```ocaml
type t
val x : t
```

If we use it as a reference interface and then parse the same interface from a different cmi, we expect our current implementation to report a modification of `x` when we would clearly consider those interfaces to be identical.

We should add this as a test and fix the implem to the treat the the two types `t` as equal across the two interfaces.

Siddhi-agg:
Fixes #44. 

I have modified the current signature to reflect the type equality for abstract types with the reference signature. I have also added an equality test case for abstract types and it works as required.

### DIFF
--- a/lib/api_watch_diff.ml
+++ b/lib/api_watch_diff.ml
@@ -22,7 +22,7 @@ let add_manifest_to_abstract_types ~ref_sig ~curr_sig =
         match item with
         | Sig_type (id, decl, _, _)
           when decl.type_kind = Type_abstract && decl.type_manifest = None ->
-            (Ident.name id, id, decl) :: acc
+            (Ident.name id, id) :: acc
         | _ -> acc)
       [] ref_sig
   in
@@ -32,8 +32,8 @@ let add_manifest_to_abstract_types ~ref_sig ~curr_sig =
       | Sig_type (id, decl, rec_st, visibility)
         when decl.type_kind = Type_abstract && decl.type_manifest = None -> (
           let name = Ident.name id in
-          match List.find_opt (fun (n, _, _) -> n = name) ref_types with
-          | Some (_, ref_id, _) ->
+          match List.find_opt (fun (n, _) -> n = name) ref_types with
+          | Some (_, ref_id) ->
               let ref_path = Path.Pident ref_id in
               let new_decl =
                 {
--- a/lib/api_watch_diff.ml
+++ b/lib/api_watch_diff.ml
@@ -26,7 +26,7 @@ let set_type_equality ref_id curr_decl =
     type_manifest = Some (Btype.newgenty (Tconstr (ref_path, [], ref Mnil)));
   }
 
-let modify_curr_sig_manifest ~ref_sig ~curr_sig =
+let set_type_equalities ~ref_sig ~curr_sig =
   let ref_abstract_types = extract_abstract_types ref_sig in
   List.map
     (fun item ->
@@ -46,7 +46,7 @@ let modify_curr_sig_manifest ~ref_sig ~curr_sig =
     curr_sig
 
 let env_setup ~ref_sig ~curr_sig =
-  let modified_curr_sig = modify_curr_sig_manifest ~ref_sig ~curr_sig in
+  let modified_curr_sig = set_type_equalities ~ref_sig ~curr_sig in
   let env = Env.empty in
   let env = Env.in_signature true env in
   let env = Env.add_signature ref_sig env in
--- a/lib/api_watch_diff.ml
+++ b/lib/api_watch_diff.ml
@@ -15,6 +15,38 @@ let env_setup ~ref_sig ~curr_sig =
   let env = Env.add_signature ref_sig env in
   Env.add_signature curr_sig env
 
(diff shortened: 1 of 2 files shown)


lib/api_watch_diff.ml
```diff
@@ -15,6 +15,38 @@ let env_setup ~ref_sig ~curr_sig =
...
+          | Some (_, ref_id) ->
+              let ref_path = Path.Pident ref_id in
+              let new_decl =
+                {
+                  decl with
+                  type_manifest =
+                    Some (Btype.newgenty (Tconstr (ref_path, [], ref Mnil)));
+                }
+              in
+              Sig_type (id, new_decl, rec_st, visibility)
```
NathanReb, line 45: I think we can extract a function here, something like:
```ocaml
val set_type_equality : type_id:Ident.t -> type_declaration -> type_declaration
```

It's doing something non trivial and wrapping this into a named function would help understand what's going on!

lib/api_watch_diff.ml
```diff
@@ -38,9 +70,12 @@ let diff_value ~typing_env ~val_name ~reference ~current =
   | exception Includecore.Dont_match _ -> Some ()
 
 let compare_values ~reference ~current =
-  let env = env_setup ~ref_sig:reference ~curr_sig:current in
+  let modified_current =
+    add_manifest_to_abstract_types ~ref_sig:reference ~curr_sig:current
+  in
```
NathanReb, line 75: This should probably be moved to the `env_setup` function since this is done specifically to setup the env properly and will be useful not only to compare values but any other item.

We don't want to compare the modified items but the original ones here.

tests/api-watch-diff/test.ml
```diff
@@ -134,14 +134,6 @@ let%expect_test "Modifying a simple type" =
   Format.printf "%a" pp_diff_list result;
   [%expect {|[Any]|}]
 
-let modify_type_in_value_signature =
```
NathanReb, line 137: Why was this removed here? I'm assuming it wasn't used anymore and that's a left over from a previous PR but I'm still curious why the CI missed it if that's indeed the case!

NathanReb (commented):
It looks great! Out of curiosity, the test was indeed failing before the fix right?

I have a few comments but again they're all about style and code clarity!

tests/api-watch-diff/test.ml
Siddhi-agg: Yes, it is indeed a leftover from #43 (I noticed it a few days ago and thought of removing it in the next PR) and even I was surprised that an unused variable wasn't caught by the CI. It may be a bug in the CI!

NathanReb: Did you get the unused value warnings when running the tests locally?

Siddhi-agg:
> It looks great! Out of curiosity, the test was indeed failing before the fix right?

Yes @NathanReb. The last test was failing before I implemented this fix. 
I also noticed that the last test was only failing when we used two different `cmi`. When I used the same variable two times, the test passed. So, it was indeed creating two signature types for `type t` with different identifiers there.

tests/api-watch-diff/test.ml
Siddhi-agg: No, I didn't.

NathanReb: Might be something about how expect_test are run. It's possible this file is never actually built in a regular way and the warnings go unnoticed.

NathanReb:
> When I used the same variable two times, the test passed. So, it was indeed creating two signature types for type t with different identifiers there.

yes that's expected, if you reuse the same signature without parsing it, the type are strictly identical. In practice that will never happen so our tests should always parse signatures separately, even if it means parsing the same string twice but we can't share the parsed cmi value.

NathanReb (commented):
I have one last comment and then we're good to go!

lib/api_watch_diff.ml
```diff
@@ -9,11 +9,48 @@ module FieldMap = Map.Make (struct
...
+
+let set_type_equality ref_id curr_decl =
+  let ref_path = Path.Pident ref_id in
+  {
+    curr_decl with
+    type_manifest = Some (Btype.newgenty (Tconstr (ref_path, [], ref Mnil)));
+  }
+
+let modify_curr_sig_manifest ~ref_sig ~curr_sig =
+  let ref_abstract_types = extract_abstract_types ref_sig in
```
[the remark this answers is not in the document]
NathanReb: Actually, the more I think about it, the more I think taking both signatures makes the most sense as it is likely that we will have to modify it further and substitute other type expression for concrete types so I think just changing the name and keeping the args as you wrote them is the way to go!

PR #50, opened 2024-06-19
Fixed implementation for concrete types (Records and Variants)

### LINKED ISSUE #45: Improve diffing of values: Concrete Types
In our ongoing effort to improve the accuracy and reliability of our interface diffing tool, we need to ensure that our implementation can correctly identify modifications to values that reference **concrete types (records or variants)**. At present, our tool may fail to detect certain modifications or incorrectly flag identical concrete types as different. 

For example:
```
type t = { a: int; b: string }
val x : t
```
If we pass this same interface given above form two different cmi, they should be treated as identical and no modification should be detected in `val x`. 

Only when we have modified the record defined in `type t` like in the interface given below, should a change in `val f` be detected.
```
type t = { a: int; b: float }
val x : t
```

Similarly for variants of a type, a change in a value which references the type should only be detected when atleast one of the variants for the type has changed.

This issue will address these shortcomings by adding relevant test cases and updating the implementation accordingly to make the added tests pass.

Siddhi-agg:
Fixes #45.

I have added test cases for concrete types which represent various possible changes in them. I have modified the implementation to set type equality between the concrete types for both the signatures.

### DIFF
--- a/lib/api_watch_diff.ml
+++ b/lib/api_watch_diff.ml
@@ -9,31 +9,12 @@ module FieldMap = Map.Make (struct
   let compare = String.compare
 end)
 
-let extract_record_types ref_sig =
+let extract_non_alias_types ref_sig =
   List.fold_left
     (fun acc item ->
       match item with
-      | Sig_type (id, { type_kind = Type_record _; _ }, _, _) ->
-          FieldMap.add (Ident.name id) id acc
-      | _ -> acc)
-    FieldMap.empty ref_sig
-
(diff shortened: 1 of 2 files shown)


NathanReb (commented):
It looks good! Glad to see we could so easily make this work without rewriting type expressions!

The tests cover more than is required for this PR but I guess we can keep them around, they'll prove useful once we deal with diffing type declarations!

NathanReb (approved):
Awesome, thanks!

PR #53, opened 2024-06-24
Added test suite for polymorphic variant types

### LINKED ISSUE #51: Diffing values referencing polymorphic variant types
We don't have tests for diffing such values and it would be good to add some, making sure we have the right behaviour regarding to those, both for inlined polymorphic variant type expression and reference to named polymorphic variant types.

We should also check that it behaves properly when changing the closed/open tags (``[> `A | `B]``/``[< `A | `B]``/``[`A | `B]``) in values.

Named polymorphic variant types should be treated as any alias type is. I expect our current implem to have the right behaviour but being sure by means of testing would be great and prevent inadvertent breakage in the future!

Siddhi-agg:
Fixes #51.

### DIFF
--- a/tests/api-watch-diff/test.ml
+++ b/tests/api-watch-diff/test.ml
@@ -434,6 +434,20 @@ let%expect_test "Inlined polymorphic variant, modified" =
   Format.printf "%a" pp_diff_list result;
   [%expect {|[Value (x, Modified)]|}]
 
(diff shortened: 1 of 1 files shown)


NathanReb (approved):
Looks great! This is quite an exhaustive suite of tests, thanks for this!

PR #55, opened 2024-06-25
Fixed implementation and Added tests for extensible variant types

### LINKED ISSUE #52: Diffing values referencing extensible variant types
This is also missing from our test suite! We should add tests to make sure values referencing such types are diffed adequatly.

They should probably be treated as regular variant types when referred to in a value description.

Here again, I expect our current implementation to do the right thing but better be safe than sorry!

Siddhi-agg:
Fixes #52.

### DIFF
--- a/lib/api_watch_diff.ml
+++ b/lib/api_watch_diff.ml
@@ -41,7 +41,8 @@ let set_type_equalities ~ref_sig ~curr_sig =
               match (ref_decl.type_kind, curr_decl.type_kind) with
               | Type_abstract, Type_abstract
               | Type_record _, Type_record _
-              | Type_variant _, Type_variant _ ->
+              | Type_variant _, Type_variant _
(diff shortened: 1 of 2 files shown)


NathanReb (approved):
Great, thanks!

NathanReb:
It makes me wonder what happens when a type changes from either one of those to another. E.g. a variant type becomes abstract.

It might be that we want to treat that equally, i.e. we report the change in the type itself but not for each occurrence of the type.

Siddhi-agg:
> It makes me wonder what happens when a type changes from either one of those to another. E.g. a variant type becomes abstract.
> 
I think that most likely should report a change in every occurence of the type as per our current implementation (since we match the `type_kind` of the types in the ref and curr signature to check if they are equal). I will add tests for that, along with some basic cases that were to be added.

PR #58, opened 2024-06-26
Added tests and Fixed implementation for type_kind changes in non-alias types

### LINKED ISSUE #56: How to deal with type_kind changes for non alias types
## Context

We consider any type without a `type_manifest` to be a non alias type, as in it's not an alias for another pre-existing type or is not exposed as such.

At the moment our implementation treats alias types as what they are, aliases rather than actual types. Meaning that if we consider the following two interfaces:
```ocaml
type t = int list
val x : t
```
and
```ocaml
type t = string list
val x : t
```
we report two modifications:
- one to the type `t` itself, as one would expect
- one to the val `x`, since if we "resolve" `t` in `val x : t`, we get two different values: `val x : int list` and `val x : string list` and therefore consider `x` has been modified

Now if we consider the following two interfaces:
```ocaml
type t = A | B
val x : t
```
and
```ocaml
type t = A | C
val x : t
```

we report only the change to the type `t`.

The reasoning here is that when a type is an alias for a type from the outside world, i.e. not defined within the library which interfaces we are diffing, such as `int list`, users can legitimately use values of this type or functions that apply to it with the "outside world". Since the type is exposed as an alias, there are other ways to produce and/or consume such values than what the library itself exposes. In the above example it is fair to assume that users could write something like:
```ocaml
let x' = List.map (fun i -> i + 1) x
```

and this code will need to be updated with the previous change.

For non alias types, i.e. types that are defined within the library it's not entirely true. Any value of type `t` is intended to be of this specific type `t` as defined by the library itself. As such we consider that if the type changes, a user will have to change some occurrences where they construct/destruct values of this type directly but fixing those will be enough to get their code to build. There is no way to get a value of type `t` from the "outside world" here, nor is there a function consuming such values that should not be updated according to the change in the type itself. I.e. there is no place where, considering a `val f : t -> int`, the fix for `f x` would be to use something else than `x` rather to fix `f` according to the type `t` changes.

## Issue

That leads to our problem here: what to do when a non alias type changes from `variant` to `record`, when it becomes `abstract` or the other way around.

Our current implementation will propagate this change to all values referencing the type, essentially treating it just as an alias type change.

It seems that we instead want to treat those the same way as a change in a non alias type without a `type_kind` change. The nature of the type remains the same, it's a local named type and therefore all we stated above still applies.

We should add tests for such changes which we currently don't have and fix the implementation to fit the desired behaviour.

Siddhi-agg:
Fixes #56.

### DIFF
--- a/lib/api_watch_diff.ml
+++ b/lib/api_watch_diff.ml
@@ -13,8 +13,8 @@ let extract_non_alias_types ref_sig =
   List.fold_left
     (fun acc item ->
       match item with
-      | Sig_type (id, ({ type_manifest = None; _ } as decl), _, _) ->
-          FieldMap.add (Ident.name id) (id, decl) acc
+      | Sig_type (id, { type_manifest = None; _ }, _, _) ->
+          FieldMap.add (Ident.name id) id acc
       | _ -> acc)
     FieldMap.empty ref_sig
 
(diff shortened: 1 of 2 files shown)


NathanReb (commented):
I think the test suite would benefit from having fewer test cases here. I think a single test case can illustrate what we're testing, maybe a couple at most, including one with a change from/to abstract.

NathanReb (approved):
Awesome, thanks!

PR #61, opened 2024-07-09
Clean up for redundant library calls

Siddhi-agg:
I have removed `Test_helpers.compile_interface` and `Api_watch_diff.diff_interface` and opened these libraries in the `test.ml` to use their functions instead.


NathanReb (approved):
I'm personally not a huge fan of global opens but here I guess it makes sense!

PR #62, opened 2024-07-11
Refactored code to improve the diffing implementation

### LINKED ISSUE #60: Improve diffing implementation
To diff the set of signatures we need to fold over a pair of maps representing each a version of the interface.

Our current implementation does by folding over both maps sequentially to be able to detect added and modified values in the first pass and removed values in the second pass.

This is not very efficient and folding over both is in fact possible using `Map.merge` which signature correspond exactly to what we want.

Siddhi-agg:
Fixes #60.

### DIFF
--- a/lib/api_watch_diff.ml
+++ b/lib/api_watch_diff.ml
@@ -81,29 +81,25 @@ let compare_values ~reference ~current =
   let env = env_setup ~ref_sig:reference ~curr_sig:current in
   let ref_values = extract_values reference in
   let curr_values = extract_values current in
-  let diffs =
-    FieldMap.fold
-      (fun val_name curr_vd acc ->
-        match FieldMap.find_opt val_name ref_values with
-        | None -> Value (val_name, Added curr_vd) :: acc
-        | Some ref_vd -> (
-            let value_differs =
-              diff_value ~typing_env:env ~val_name ~reference:ref_vd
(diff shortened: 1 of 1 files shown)


lib/api_watch_diff.ml
```diff
@@ -81,29 +81,25 @@ let compare_values ~reference ~current =
...
+              ~current:curr_vd
+          in
+          match value_differs with
+          | None -> None
+          | Some _ ->
+              Some
+                (Value (val_name, Modified { ref = ref_vd; current = curr_vd }))
+          ))
+    ref_values curr_values
+  |> FieldMap.bindings |> List.map snd
```
NathanReb, line 102: You can use `Seq` here to avoid allocating lists that we're not going to use.
See https://ocaml.org/manual/5.2/api/Seq.html for the documentation but here you can turn this into:
```ocaml
|> FieldMap.to_seq |> Seq.map_snd |> List.of_seq
```
It might be a bit overkill here but it's good to know it's a possibility!

NathanReb (approved):

NathanReb:
It looks so much better that way!

lib/api_watch_diff.ml
Siddhi-agg: Will look into that too, it may prove useful in the future!

PR #63, opened 2024-07-11
Added failing test for identical values 

Siddhi-agg:
This is the behaviour I was talking about. For values referencing types with constructors, our implementation isn't working as expected.

### DIFF
--- a/tests/api-watch-diff/test.ml
+++ b/tests/api-watch-diff/test.ml
@@ -643,7 +643,7 @@ let%expect_test "Changing from record to variant type" =
   Format.printf "%a" pp_diff_list result;
   [%expect {|[Any]|}]
 
-let%expect_test "New test: Identical values" =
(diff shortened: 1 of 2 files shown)


tests/api-watch-diff/test.ml
```diff
@@ -661,3 +661,22 @@ let%expect_test "New test: Identical values" =
   let result = diff_interface ~reference ~current in
   Format.printf "%a" pp_diff_list result;
   [%expect {|[]|}]
+
+let%expect_test " Values referencing types with constructors, modified" =
```
NathanReb, line 665: I think you mean parameters here and not constructors:
```suggestion
let%expect_test " Values referencing types with parameters, modified" =
```

NathanReb (commented):
Besides the test name it looks good! Thanks for finding and fixing this bug!

NathanReb (approved):
Thanks! Let's merge!

PR #64, opened 2024-07-15
Used ocaml-diffutils for printing text diffs

Siddhi-agg:
This PR makes the following changes:
- Vendored `ocaml-diffutils` to be used in this project for printing text diffs.
- Created a `to_diff` function in `api_watch_diff.ml` to convert `diff list` to `Diffutils.Diff.t`.
- Modified `api-diff.ml` to print text diffs for the cram tests


bin/api_diff.ml
```diff
@@ -1,3 +1,5 @@
+open Diffutils
```
NathanReb, line 1: I'd prefer not to use a global open here. You can eventually use a local one where it makes sense:
```ocaml
let open Diffutils in
...
```

bin/api_diff.ml
```diff
@@ -6,7 +8,13 @@ let run (`Ref_cmi reference) (`Current_cmi current) =
       ~current:current.cmi_sign
   in
   if differs = [] then Printf.printf "API unchanged!\n"
-  else Printf.printf "API changed!\n"
+  else
+    match differs with
```
NathanReb, line 12: Probably good to rename `differs` which was named that way when it was a boolean to `diff`.

bin/api_diff.ml
```diff
@@ -6,7 +8,13 @@ let run (`Ref_cmi reference) (`Current_cmi current) =
...
+    | Api_watch_diff.Any :: _ -> Printf.printf "API changed!\n"
```
NathanReb, line 13: I'd suggest dropping this header entirely and have the tool behave like `diff`:
1. exit with status 0 if there is no diff, no output
2. exit with status 1 if there is a diff, printing the diff

For `Any`, we already transform it into a textual diff so no need to have a specific treatment here.

NathanReb: To deal with the exit status you can use `Cmdliner.Cmd.eval'` instead of the `Cmdliner.Cmd.eval` that we use and make our run function return the appropriate exit status instead of `()`.

lib/api_watch_diff.ml
```diff
@@ -1,4 +1,5 @@
 open Types
+open Diffutils.Diff
```
NathanReb, line 2: Same remark as above, this is only relevant in the `to_diff` function.

lib/api_watch_diff.ml
```diff
@@ -120,3 +121,34 @@ let diff_interface ~reference ~current =
     | _, _ -> [ Any ]
     | exception Includemod.Error _ -> [ Any ]
   else value_diffs
+
+let to_diff (diff_result : diff list) : t =
```
NathanReb, line 125: I'd use `to_text_diff` here. We have our own `diff` type so this name is a bit confusing.

lib/api_watch_diff.ml
```diff
@@ -120,3 +121,34 @@ let diff_interface ~reference ~current =
...
+            {
+              orig = [ "<unspecified change>" ];
+              new_ = [ "<unspecified change>" ];
+            }
+      | Value (name, change) ->
+          let diff =
+            match change with
+            | Added vd -> { orig = []; new_ = [ vd_to_string name vd ] }
+            | Removed vd -> { orig = [ vd_to_string name vd ]; new_ = [] }
+            | Modified { ref; current } ->
```
NathanReb, line 147: I just realized we're using `ref` which is also used to create a mutable value in OCaml. We might want to rename it `ref_` in a separate PR!

lib/api_watch_diff.ml
```diff
@@ -120,3 +121,34 @@ let diff_interface ~reference ~current =
...
+    Buffer.contents buf
+  in
+  List.map
+    (fun item ->
+      match item with
+      | Any ->
+          Diff
+            {
+              orig = [ "<unspecified change>" ];
+              new_ = [ "<unspecified change>" ];
```
NathanReb, line 140: Id' suggest going for:
```suggestion
              orig = [];
              new_ = [ "<unsupported change>" ];
```
This will make a more readable output and using `unsupported` instead of `unspecified` makes it clearer to the user that this is something to be fixed on our end.

NathanReb (commented):
I was not aware that diffutils was not released to opam! @panglesd any chance we can get a release?

In the meantime I'd rather avoid vendoring here as it is not required. We can instead use opam `pin-depends` (see https://opam.ocaml.org/doc/Manual.html#opamfield-pin-depends). You will have to both add diffutils to our dependencies and add the proper `pin-depends` field. You'll want to use an opam git URL, probably:
```
git+https://github.com/panglesd/ocaml-diffutils#0fc293bc0cd2c34742552f7a0c777bfba385bd90
```
We could have used `#main` but using a fixed commit here prevents involuntary updates when something is pushed to the main branch.

api-watch.opam
```diff
@@ -13,6 +12,10 @@ depends: [
   "logs"
   "cmdliner" {>= "1.1.0"}
   "odoc" {with-doc}
+  "diffutils"
+]
+pin-depends: [
```
NathanReb, line 17: I forgot to mention it but our opam file is generated by dune so we don't edit it directly. See https://dune.readthedocs.io/en/latest/howto/opam-file-generation.html for opam generation documentation.

You should instead edit the `dune-project` file. You'll be able to add diffutils to the list of dependencies but you won't be able to add a pin-depends there. To add pin-depends we'll instead have to add an `api-watch.opam.template` file to the repo and set the fields that can't be set in `package` stanza of our `dune-project` there.

api-watch.opam.template
```diff
@@ -0,0 +1,34 @@
+opam-version: "2.0"
```
NathanReb, line 1: The template file only needs to define the `pin-depends` field. The other fields will be generated by dune.

The way opam.template files work with dune is that if it finds a field defined in the template it will use the field as it is in the template and will generate it otherwise. Here it basically overrides the whole generation process.

panglesd:
> @panglesd any chance we can get a release?

Non zero chances, but I'd like at some point to rethink the interface with a fresh mind! (since I've not touched it in quite a long time).

I'd say: go with the pin-depends, and I won't forget to ping you whenever I have made the release!

NathanReb:
I think you can undraft @Siddhi-agg !

NathanReb (approved):
Looks good, thanks!

Let's merge this!

PR #66, opened 2024-07-18
Added implementation to diff modules and print their text diffs 

### LINKED ISSUE #65: Add support for diffing submodules
At the moment the only items we diff ourselves (i.e. not through `Includemod.signatures`) are `value_description`s.

We'd like to add support for diffing `module_declaration`s as it's one of the simplest ones to add since it should roughly amount to a recursive call to our current `diff_interface`.
That means that we will treat those modules the same way we treat the main one, that is, we diff value_descriptions (and modules) in them, and use `Includemod.signatures` if we couldn't detect a change that way.

## Diffing modules

We'll need to make our diff type recursive as well to be able to represent at least module modifications, not sure whether we want to represent the addition or removal of a module as the addition or removal of it's whole content.

It should look something along those lines:
```ocaml
type 'a change =
  | Added of 'a
  | Removed of 'a
  | Modified of {ref_: 'a; current: 'a}

type module_diff =
  { module_name : Ident.t
  ; changes : module_change
  }
and item_change =
  | Value of { name : Ident.t; change : value_description change }
  | Module of module_diff
and module_change =
  | Unsupported
  | Supported of item_change list
```

We could then change our entry point `diff_interface` to return a `module_diff` option, where `None` means no change.

Starting from this point, even changes in the main module would be "namespaced" as changes to the main module.

The end goal would be to ultimately drop the intermediate `module_change` type that only serves as a way to report changes that we do not detect ourselves yet.

## Converting to text diff

We'll now also need a good way to turn this into a text diff as it won't be as easy to represent as it is now.

One approach would be to have a map from fully qualified module paths (e.g. `M`, `M.M2` or `M.M2.M3`) to `Diffutils.Diff.t`. This not necessarily the perfect approach but has the advantage to be fairly simple to build and non recursive while still allowing to use `Diffutils` printing utilities.

We'd then need to prefix the printing of diffs with the module they belong to:
```
diff module M
- val f : int -> int
+ val f : int -> (int, string) result
diff module M.M2
- val x : int
...
```

Siddhi-agg:
Fixes #65.

### DIFF
--- a/lib/api_watch_diff.ml
+++ b/lib/api_watch_diff.ml
@@ -204,7 +204,10 @@ let to_text_diff (diff_result : module_diff) : Diffutils.Diff.t String_map.t =
             match change with
             | Value { name; change = val_change } ->
                 let diff = process_value_diff name val_change in
-                String_map.add module_path diff acc'
+                String_map.update module_path
+                  (function
+                    | None -> Some diff | Some existing -> Some (existing @ diff))
+                  acc'
             | Module sub_module_diff ->
                 let sub_module_path =
                   module_path ^ "." ^ sub_module_diff.module_name
--- a/lib/api_watch_diff.ml
+++ b/lib/api_watch_diff.ml
@@ -5,20 +5,15 @@ type 'item change =
   | Removed of 'item
   | Modified of { ref_ : 'item; current : 'item }
 
-type module_diff = {
-  module_name : string;
-  changes : module_change;
-}
+type module_diff = { module_name : string; changes : module_change }
 
 and item_change =
   | Value of { name : string; change : value_description change }
   | Module of module_diff
 
-and module_change =
-  | Unsupported
-  | Supported of item_change list
+and module_change = Unsupported | Supported of item_change list
 
-type diff = module_diff option 
+type diff = module_diff option
 
 module FieldMap = Map.Make (struct
   type t = string
@@ -76,7 +71,8 @@ let extract_modules items =
   List.fold_left
     (fun tbl item ->
       match item with
-      | Sig_module (id, _, mod_decl, _, _) -> FieldMap.add (Ident.name id) mod_decl tbl
+      | Sig_module (id, _, mod_decl, _, _) ->
+          FieldMap.add (Ident.name id) mod_decl tbl
       | _ -> tbl)
     FieldMap.empty items
 
@@ -110,8 +106,10 @@ let compare_values ~reference ~current =
     (fun val_name ref_opt curr_opt ->
       match (ref_opt, curr_opt) with
       | None, None -> None
-      | None, Some curr_vd -> Some (Value {name=val_name; change= Added curr_vd})
-      | Some ref_vd, None -> Some (Value {name=val_name; change= Removed ref_vd})
+      | None, Some curr_vd ->
+          Some (Value { name = val_name; change = Added curr_vd })
+      | Some ref_vd, None ->
+          Some (Value { name = val_name; change = Removed ref_vd })
       | Some ref_vd, Some curr_vd -> (
           let value_differs =
             diff_value ~typing_env:env ~val_name ~reference:ref_vd
@@ -120,8 +118,12 @@ let compare_values ~reference ~current =
           match value_differs with
           | None -> None
           | Some _ ->
-              Some (Value {name=val_name; change=Modified {ref_=ref_vd; current=curr_vd}})
-          ))
+              Some
+                (Value
+                   {
+                     name = val_name;
+                     change = Modified { ref_ = ref_vd; current = curr_vd };
+                   })))
     ref_values curr_values
   |> FieldMap.bindings |> List.map snd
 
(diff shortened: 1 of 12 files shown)


NathanReb (commented):
It looks good and seems to work well!

The code can be simplified quite a bit, I left a few comments in that regard.

It would also be good to add an expect test with a submodule!

lib/api_watch_diff.ml
```diff
@@ -59,6 +67,15 @@ let env_setup ~ref_sig ~curr_sig =
   let env = Env.add_signature ref_sig env in
   Env.add_signature modified_curr_sig env
 
+let extract_modules items =
```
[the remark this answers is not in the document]
Siddhi-agg: Should we do that in this PR itself? Or a separate one?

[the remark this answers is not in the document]
NathanReb: It should definitely be done here as it will greatly simplify the PR I think!

lib/api_watch_diff.mli
```diff
@@ -1,13 +1,23 @@
...
+module FieldMap : Map.S with type key = string
 
 val diff_interface :
   reference:Types.signature_item list ->
   current:Types.signature_item list ->
-  diff list
+  diff
 
-val to_text_diff : diff list -> Diffutils.Diff.t
+val to_text_diff : diff -> Diffutils.Diff.t FieldMap.t
```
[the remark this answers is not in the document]
Siddhi-agg: Help to write docs would be appreciated!

Siddhi-agg:
> It would also be good to add an expect test with a submodule!

We already have expect tests for adding, removing and modifying submodules of the main module. Can you clarify what other kind of submodule test should we add?

lib/api_watch_diff.mli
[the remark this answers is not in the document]
NathanReb: Sure, let me know when you're done with the other comments and I'll push some doc on this branch.

lib/api_watch_diff.ml
```diff
@@ -81,69 +96,123 @@ let diff_value ~typing_env ~val_name ~reference ~current =
...
-    let formatter = Format.formatter_of_buffer buf in
-    Printtyp.value_description (Ident.create_local name) formatter vd;
-    Format.pp_print_flush formatter ();
-    Buffer.contents buf
+  let ref_items = extract_items reference in
+  let curr_items = extract_items current in
+  let item_changes =
+    Sig_item_map.merge
+      (fun (item_type, name) ref_opt curr_opt ->
+        match (item_type, ref_opt, curr_opt) with
```
NathanReb, line 106: There are a lot of things to simplify here.

I noticed that `diff_value` still returns a `unit option`. That's a legacy from a time before we had a representation for diffs. We should make it return a `value_description change option` at the very least. The point of making it return a `unit option` initially was so that we could replace the `unit` in there by a diff type for value descriptions.

You can make this pattern-matching much simpler to read by flattening it, there's not need for nested pattern matching here. For example the value case can be turned into something like:
```ocaml
| Value_item, None, None -> None
| Value_item, Some ref, None -> Some (Value {name; change = Removed ref})
| Value item, None, Some curr -> Some (Value {name; change = Added ref})
| Value_item, Some ref, Some curr ->
  let change = diff_value ~typing_env ... in
  Option.map (fun change -> Value {name; change}) change
 | Module_item, None, None, -> None
  ...
```

Alternatively, you could extract a `diff_value_item` function with the following signature:
```ocaml
val diff_value_item : name: string -> ref: value_description option -> curr: value_description option -> item_change option
```
so that the diff_items function just has to dispatch based on item_type and call other functions to keep things simple.

lib/api_watch_diff.ml
```diff
@@ -81,69 +96,123 @@ let diff_value ~typing_env ~val_name ~reference ~current =
...
+            | None, Some (Mod _curr_md) ->
+                Some (Module { module_name = name; changes = Unsupported })
+            | Some (Mod _ref_md), None ->
+                Some (Module { module_name = name; changes = Unsupported })
+            | Some (Mod ref_md), Some (Mod curr_md) -> (
+                match (ref_md.md_type, curr_md.md_type) with
+                | Mty_signature ref_submod, Mty_signature curr_submod ->
+                    diff_items name ~reference:ref_submod ~current:curr_submod
+                | _, _ ->
+                    Some (Module { module_name = name; changes = Unsupported }))
```
NathanReb, line 142: This is incorrect, you can't state that there will necessarily be a change there.

It's okay to stick to simple module declaration as you did above but we should default to calling some `Includemod.signatures` (probably `Includemod.modtypes`) like function here to detect changes when we don't know how to diff those.

NathanReb (commented):
I wasn't thorough enough in my previous review and missed a few things.

I think the recursion is not properly handled for modules and the submodules are not diffed the same way as the entry module is (c.f. my comment on `Includemod.signatures` only being called in `diff_interface`) when they should.

I did miss the fact that we would have to deal with functors, which should be easy enough to deal with but we can leave that for a separate PR!

Alias and ident modtypes are a bit different as those probably fall into things we want to translate into concrete signatures before diffing but we still need proper handling of those in the meantime.

This are all things for which we should have tests included in this PR.

All this means we need a better way to represent module diffs as it can't be restricted to a signature diff because we need to cover those other cases, we need an extra layer to represent `modtypes` diff.

Here are what I think we need to get this PR to a mergeable state:
1. Add tests, there are a lot of cases we don't check for here and that we should cover, e.g. if a module declaration changes its modtype, like from regular module decl to functor, or to alias. Another test is the one I mention in the comments, where we check that we properly call `Includemod.signatures` if needed.
2. Fix the recursion for regular modules so we call `Includemod.signatures` when we don't detect a change with our diffing algorithm.
3. Add a layer to deal with modtypes, both in the diff types and the functions we have

tests/api-diff/run.t
```diff
@@ -207,6 +215,58 @@ Compile the new .mli file to a .cmi file
...
+  -val f : int -> string
+  +val f : int -> (string, string) result
+  
+  diff module Modified_module.D:
+  -val b : int list -> int
+  +val b : float list -> float
+  -val g : int -> string
+  +val g : int -> (string, string) result
+  
+  diff module Modified_module.E:
```
NathanReb, line 269: This should be marked as a module addition shouldn't it?

tests/api-watch-diff/test_module.ml
```diff
@@ -0,0 +1,141 @@
...
+    module N : sig val x: int end
+  |}
+  in
+  let result = diff_interface ~module_name:"Main" ~reference ~current in
+  Format.printf "%a" pp_diff_list result;
+  [%expect
+    {|
+    Some Module Main: [ Value (a, Modified); Value (f, Modified);
+    Module M: [ Value (b, Modified); Value (g, Modified);]
+    Module N: Unsupported changes]|}]
```
NathanReb, line 66: Same here, I think we should still be able to detect added and removed modules at least. It's ok to report unsupported changes when two versions of a module are not both `Mty_signatures` but those simple changes we should be able to detect.

NathanReb (commented):
It looks good! Can you add support for modules addition and removal and then we should be good to go!

lib/api_watch_diff.ml
```diff
@@ -11,7 +11,11 @@ and item_change =
   | Value of { name : string; change : value_description change }
   | Module of module_diff
 
-and module_change = Unsupported | Supported of item_change list
+and module_change =
+  | Unsupported
+  | Supported of item_change list
+  | Mod_added of module_declaration
+  | Mod_removed of module_declaration
```
NathanReb, line 18: How about instead reworking the `'item change` type. I think it's about the right time to do this.

We could instead define the following types:
```ocaml
type ('item, 'diff) diff =
  | Added of 'item
  | Removed of 'item
  | Modified of 'diff
  
 (** The simplest diff representation for the modification of a value of type 'a.
     [reference] is the value before and [current] is the value after the change occured.
     Use this type when there is no better representation available. *)
 type 'a atomic_modification =
   { reference : 'a
    ; current : 'a
    }
  
  type value_diff =
    { vname : string
    ; vdiff : (value_description, value_description atomic_modification) diff
    }
  
  type module_diff =
    { mname : string
    ; mdiff : (module_declaration, module_modification) diff
    }
  and module_modification =
    | Unsupported
    | Supported of item_diff list 
  and item_diff =
    | Value of value_diff
    | Module of module_diff
 ```
 
 This uses more consistent naming and structure and allows for better representation. What do you think?

Siddhi-agg: Yes, this definitely has better structure than the current implementation! It can also be used better when we add support for more `signature_items`.

NathanReb (approved):
Looks great! Let's merge!

lib/api_watch_diff.mli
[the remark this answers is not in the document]
NathanReb: I'll deal with this when splitting out modules!

PR #67, opened 2024-08-16
Upgraded to OCaml 5.2 and OCamlformat 0.26.2

Siddhi-agg:
This PR upgrades the project to use OCaml version 5.2 instead of 5.1, and also upgrades OCamlformat version from 0.26.1 to 0.26.2 to make it compatible with the upgraded compiler.


NathanReb (approved):
Looks good, thanks!

PR #68, opened 2024-08-16
Added implementation to get a merged signature for an entire dune built library

Siddhi-agg:
This PR makes the following changes:

- Added a function `dune_library_to_signature` which takes the path of a dune built library and extracts the signatures of all the different module files in that library, to build a "merged" signature of the entire library
- Modified the `api_diff.ml` to now either take two `.cmi` files (this uses the only the diffing implementation) OR two library paths (this uses the `dune_library_to_signature` and then the diffing implementation) as reference and current versions and then diff them
- Added a cram test which creates two versions (the second one with some modifications to test their diff) of a simple OCaml library with one main `math.ml` and another `utils.ml` file and uses `api-diff` command to create their text diff

**Note:**
The implementation of `dune_library_to_signature` requires that the project/library whose path is provided has been properly built and a `_build` folder is present for it in the path provided.

### DIFF
--- a/bin/api_diff.ml
+++ b/bin/api_diff.ml
@@ -1,13 +1,19 @@
 let run (`Ref_cmi reference) (`Current_cmi current) =
-  let reference_sig, current_sig, module_name =
+  let open CCResult.Infix in
+  let* reference_sig, current_sig, module_name =
     if Sys.is_directory reference && Sys.is_directory current then
-      let reference_sig = Api_watch.Library.load reference in
-      let current_sig = Api_watch.Library.load current in
+      let+ reference_sig = Api_watch.Library.load reference
+      and+ current_sig = Api_watch.Library.load current in
       let module_name = Filename.basename current in
       (reference_sig, current_sig, module_name)
     else
-      let reference_cmi = Cmi_format.read_cmi reference in
-      let current_cmi = Cmi_format.read_cmi current in
+      let+ reference_cmi =
+        try Ok (Cmi_format.read_cmi reference)
+        with e -> Error (Printexc.to_string e)
+      and+ current_cmi =
+        try Ok (Cmi_format.read_cmi current)
+        with e -> Error (Printexc.to_string e)
+      in
       let module_name = current_cmi.cmi_name in
       (reference_cmi.cmi_sign, current_cmi.cmi_sign, module_name)
   in
@@ -16,7 +22,7 @@ let run (`Ref_cmi reference) (`Current_cmi current) =
       ~current:current_sig
   in
   match diff with
-  | None -> 0
+  | None -> Ok 0
   | Some diff ->
       let text_diff = Api_watch.Text_diff.from_diff diff in
       let print_module_diff module_path diff =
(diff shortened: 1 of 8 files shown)


NathanReb (commented):
This looks good!

There are quite a few things to improve for it to work properly with dune built libraries, in particular here it shows module names from dune compilation units rather than the module names that the users will reference.
I think we can fix this in subsequent PRs though.

I added a few comments, especially about error handling. To fix this you'll have to carry `result` values around which can be cumbersome without a few tricks.

I was about to suggest using https://github.com/yawaramin/letops but it's not been released to opam unfortunately. We can define our own local `Letops` module with a `Result` submodule to help with that. If you want to learn a bit more about those binding operators and how to use them you can take a look at the OCaml manual relevant section here: https://ocaml.org/manual/5.2/bindingops.html. I'm of course happy to walk you through it as well if you'd like!

bin/api_diff.ml
```diff
@@ -38,6 +54,13 @@ let info =
   Cmd.info "api-watcher" ~version:"%%VERSION%%" ~exits:Cmd.Exit.defaults
     ~doc:"List API changes between two versions of a library"
 
-let term = Cmdliner.Term.(const run $ ref_cmi $ current_cmi)
+let run_with_error_handling ref_cmi current_cmi =
+  match run ref_cmi current_cmi with
+  | Ok code -> code
+  | Error msg ->
+      Printf.eprintf "Error: %s\n" msg;
+      2
```
NathanReb, line 62: You can use this `Cmdliner` function in place of `eval`: https://erratique.ch/software/cmdliner/doc/Cmdliner/Cmd/index.html#val-eval_result'

That way you can build the cmdliner term with `run`.

bin/api_diff.ml
```diff
@@ -1,12 +1,28 @@
...
+  let* reference_sig, current_sig, module_name =
+    if Sys.is_directory reference && Sys.is_directory current then
+      let+ reference_sig = Api_watch.Library.load reference
+      and+ current_sig = Api_watch.Library.load current in
+      let module_name = Filename.basename current in
+      (reference_sig, current_sig, module_name)
+    else
+      let+ reference_cmi =
+        try Ok (Cmi_format.read_cmi reference)
+        with e -> Error (Printexc.to_string e)
```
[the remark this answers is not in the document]
NathanReb: It's probably a good idea to expose the one we have in `Library` for now!

lib/library.ml
```diff
@@ -0,0 +1,64 @@
...
+    try Ok (Sys.readdir dir)
+    with Sys_error e ->
+      Error (Printf.sprintf "Error reading directory %s: %s" dir e)
+  in
+  Array.fold_left
+    (fun acc_res file ->
+      let* acc = acc_res in
+      let path = Filename.concat dir file in
+      try
+        if Sys.is_directory path then
```
NathanReb, line 13: This should ignore directories and only load `.cmi` files from the given directory. The idea is that users should call this function on, for instance, `_build/install/default/lib/re` or `_opam/lib/re`.

I.e. let's keep this as simple as possible and not bother with lib discovery logic for now. The current implem is likely to load a lot of irrelevant modules.

NathanReb (commented):
The new error handling looks good, thanks!

Still have a few comments but then we should be good to go!

NathanReb (approved):

PR #73, opened 2024-08-27
Fix text diff representation of changes that span over multiple lines

NathanReb:
Diffs were previously not properly converted to text diffs when an individual value or module change would span over several lines.

It was turned into a single line of diff but was pretty printed over multiple lines, ending up in a misrepresented diff.

This adds a regression tests and a subsequent fix.

To simplify all this I extracted a `Text_diff.pp` function which I used to write an expect tests.


Siddhi-agg (approved):
This looks great!

PR #74, opened 2024-08-27
Explicitly pass the main module name through the command line

NathanReb:
This used to rely on some inference based on parent directories which was unpractical and unreliable.

### DIFF
--- a/bin/api_diff.ml
+++ b/bin/api_diff.ml
@@ -1,24 +1,31 @@
-let run (`Ref_cmi reference) (`Current_cmi current) =
+let run (`Main_module main_module) (`Ref_cmi reference) (`Current_cmi current) =
   let open CCResult.Infix in
   let* reference_sig, current_sig, module_name =
-    if Sys.is_directory reference && Sys.is_directory current then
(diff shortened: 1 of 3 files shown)


bin/api_diff.ml
```diff
@@ -1,24 +1,31 @@
...
+        let module_name = String.capitalize_ascii main_module in
+        (reference_sig, current_sig, module_name)
+    | false, false, main_module ->
+        let () =
+          match main_module with
+          | None -> ()
+          | Some _ ->
+              Printf.eprintf
+                "%s: --main-module ignored when diffing single .cmi files"
+                Sys.executable_name
```
Siddhi-agg, line 19: I think we can also add a test for this (when both files are simple .cmi and a `--main-module` is also provided to the terminal) as a part of the cram tests.

Siddhi-agg (commented):
Other than this comment, the PR looks great!

bin/api_diff.ml
NathanReb: You're right, I'll add a test!

PR #86, opened 2024-10-02
Split the monolithic cram test into multiple files

### LINKED ISSUE #85: Split the monolithic cram test into multiple files
The [`tests/api-diff/run.t`](https://github.com/ocaml-semver/ocaml-api-watch/blob/main/tests/api-diff/run.t) file contains tests for the `api-diff` tool, "end to end": that is, checking that some input to the tool produces the expected final output.

The tests are written using [`dune`'s support from cram tests.](https://dune.readthedocs.io/en/stable/tests.html#cram-tests)

However, almost all cram tests are in a single file. This makes it harder to read, and won't scale when we add more tests.

It would be better to group tests by features, in separate files with a relevant name.

Dezzy12:
fixes #85 
I have splitted the `run.t` file into the following:
- identical_cmi.t 
- module_tests.t  
- project_comparison.t  
- type_tests.t  
- value_tests.t

@panglesd


panglesd:
Hello @Dezzy12 and thanks for the PR!

Unfortunately, the tests are not passing anymore, as you can see from the result of the CI (the red cross next to "`ocaml-ci`", you can click on "details" to see what went wrong).

To test locally, you can run `dune runtest`. This will show you the difference between what is expected and what is the result of the test.
If you want to change the expected value to what running the test produces, you can run `dune promote`. `dune runtest --auto-promote` is a shorthand for `dune runtest` followed by `dune promote`.

You can have more information about tests in the [dune doc on writing and running tests](https://dune.readthedocs.io/en/stable/tests.html), in particular its [section on cram tests](https://dune.readthedocs.io/en/stable/tests.html#cram-tests).

Dezzy12:
I have fixed the issue. Sorry for the Delay
@panglesd

tests/api-diff/type_tests.t
```diff
@@ -0,0 +1,78 @@
...
+  > EOF
+
+We generate the .cmi file
+
+  $ ocamlc ref.mli
+
+And now we run api-watcher on that same cmi file as both arguments,
+there should be no diff:
+
+  $ api-diff ref.cmi ref.cmi
```
panglesd, line 16: This test is done twice, in `type_tests` and in `value_tests`.

I think it makes sense to extract those "diffs between identical cmis" in a specific file.

panglesd (approved):
Apart from the small comment, looks good to me!

tests/api-diff/type_tests.t
panglesd: Looking at the history, you had the tests extracted before! You removed that when adding the `ref.mli` creation to the other tests using the file.

It is okay to create the same `ref.mli` file in different tests. But do bring back your `identical_cmi.t` file!

Dezzy12:
I have implemented the requested changes

panglesd:
Thanks!

PR #87, opened 2024-10-03
Added support for diffing class_type_declaration items

RitaAkor:
Hello!
I apologize for the delay in submitting the update. My build tests kept failing, and it took me some time to resolve the issues.

This adds detection for addition and removal of class_type_declaration items in the API diffing process. 

Changes include:

- Extended `item_type` and `sig_items` types in `lib/diff.ml` to include class_type_declaration.
- Updated `extract_items` function to handle class_type_declaration items.
- Modified `Diff.items` function to compare class_type_declaration items.
- Added tests in `tests/api-watch/test_diff_class_type_decl.ml` to verify the new functionality.

Thank you!

### DIFF
--- a/lib/diff.ml
+++ b/lib/diff.ml
@@ -21,7 +21,11 @@ and module_modification = Unsupported | Supported of sig_item list
 and sig_item = Value of value | Module of module_
 
 type item_type = Value_item | Module_item | ClassType_item [@@deriving ord]
-type sig_items = Val of value_description | Mod of module_declaration | Cty of class_type_declaration (*Extended for class-type-dec*)
+
+type sig_items =
+  | Val of value_description
+  | Mod of module_declaration
+  | Cty of class_type_declaration (*Extended for class-type-dec*)
 
 module Sig_item_map = Map.Make (struct
   type t = item_type * string [@@deriving ord]
@@ -35,8 +39,11 @@ let extract_items items =
           Sig_item_map.add (Module_item, Ident.name id) (Mod mod_decl) tbl
       | Sig_value (id, val_des, _) ->
           Sig_item_map.add (Value_item, Ident.name id) (Val val_des) tbl
-      | Sig_class_type (id, class_type_decl, _, _) ->  (*Extended to include Class_type_declaration*)
-          Sig_item_map.add (ClassType_item, Ident.name id) (Cty class_type_decl) tbl
+      | Sig_class_type (id, class_type_decl, _, _) ->
+          (*Extended to include Class_type_declaration*)
+          Sig_item_map.add
+            (ClassType_item, Ident.name id)
+            (Cty class_type_decl) tbl
       | _ -> tbl)
     Sig_item_map.empty items
 
@@ -74,7 +81,8 @@ let value_item ~typing_env ~name ~reference ~current =
       | _, _ ->
           Some (Value { vname = name; vdiff = Modified { reference; current } })
       | exception Includecore.Dont_match _ ->
-          Some (Value { vname = name; vdiff = Modified { reference; current } }))
+          Some (Value { vname = name; vdiff = Modified { reference; current } })
+      )
   | _ -> None
 
 let rec items ~reference ~current =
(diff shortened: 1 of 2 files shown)


panglesd:
Hello @RitaAkor!

Thanks for the contribution, it looks great! I'll have a deeper look today. In the meantime, could you add some "cram tests"? It is always useful to test the feature in a real scenario.

> I apologize for the delay in submitting the update.

On the contrary, that was pretty fast! In my experience, programming always take more time than planned.

RitaAkor:
Hi @panglesd 

Thank you so much for the feedback! I’ll go ahead and add some cram tests.

lib/diff.ml
```diff
@@ -86,7 +95,12 @@ let rec items ~reference ~current =
       | Value_item, reference, current ->
           value_item ~typing_env:env ~name ~reference ~current
       | Module_item, reference, current ->
-          module_item ~typing_env:env ~name ~reference ~current)
+          module_item ~typing_env:env ~name ~reference ~current
+      | ClassType_item, Some (Cty _), None ->
+          None (* Class type was removed, ignore it *)
+      | ClassType_item, None, Some (Cty _) ->
+          None (* Class type was added, ignore it *)
+      | ClassType_item, _, _ -> None)
```
NathanReb, line 103: First, could you please move the pattern matching over the reference vs current binding options to a separate function similarly to how it's done for the other item types? This avoid cluttering this function with any logic and let it's sole role to dispatch the diffing based on the item type.

This implementation is unfortunately incorrect. In the cases where a class type was added or removed, you should be returning something. You can look at how the `value_item` function is implemented to take inspiration from it.

tests/api-watch/test_diff_class_type_decl.ml
```diff
@@ -0,0 +1,95 @@
...
+        clty_loc = Location.none;
+        clty_attributes = [];
+        clty_uid = Uid.internal_not_actually_unique;
+      }
+
+    let signature_with_class_type decl =
+      [
+        Sig_class_type
+          (Ident.create_local "test_class", decl, Trec_first, Exported);
+      ]
```
NathanReb, line 45: You can use our `Test_helpers` library instead to get the compiled interface from OCaml sources. 

The function you want here is `Test_helpers.compile_interface`.

You can take a look at some existing expect tests we have in `test_diff.ml` or `test_diff_module.ml` to see an example of how it's used.

lib/diff.ml
```diff
@@ -35,6 +39,11 @@ let extract_items items =
           Sig_item_map.add (Module_item, Ident.name id) (Mod mod_decl) tbl
       | Sig_value (id, val_des, _) ->
           Sig_item_map.add (Value_item, Ident.name id) (Val val_des) tbl
+      | Sig_class_type (id, class_type_decl, _, _) ->
+          (*Extended to include Class_type_declaration*)
```
NathanReb, line 43: Can you please remove this comment as it does not bring useful information and will age poorly.

NathanReb (commented):
This looks promising! You are correctly collecting classes in the `Sig_item_map` but the diffing part is wrong as it returns no diff no matter what happens with those items.

lib/diff.ml
RitaAkor: Thanks for the feedback! I will work on  modeling the diffing logic after value_item.Could you confirm if the structure for class type diffs should mirror exactly as value_item, or if there are subtle differences I should consider?

panglesd: Diffing of values currently detect "addition of a value", "removal of a value", and "modification of a value".

In the case of #83, the goal is to detect the additions and removal of `class_type_declarations`. It is not required to detect modifications of class types (that will come later!).

So no, it does not mirror _exactly_ `value_item`, it should be simpler! Try to understand the `value_item` implementation, to be able to take inspiration from it.

azzsal:
This PR has been opened for a long period of time and hasn't been updated in a while. @RitaAkor if you need help or have a specific question I can help out! If you're no longer interested in, I can pick it up and continue working on this. @panglesd @NathanReb.

NathanReb:
I'm closing this as we haven't heard about @RitaAkor since they were last pinged! Please feel free to comment if you'd like to resume working on this.

PR #88, opened 2024-10-03
Dropped diffutils dependency

### LINKED ISSUE #79: Drop diffutils dependency
We are currently using the [`diffutils` project](https://github.com/panglesd/ocaml-diffutils) to print API diffs as human readable text diffs. `diffutils` is not released to opam and it prevents us from releasing `api-watch`.

Vendoring the whole diffutils project would be a bit overkill as we're only using a very small part of it.

We would like to instead embed the types and functions we use directly in the project.

This makes sense as it is likely that our need for custom diff printing utilities will grow as we move toward a proper structural diff representation of API diffs.

## What to import

All of our uses of `Diffutils` are contained within our [`Text_diff` module](https://github.com/ocaml-semver/ocaml-api-watch/blob/main/lib/text_diff.ml). All the types we use come from `Diffutils.Diff`:
- the type `t` that's used to represent a textual diff
- the `printer` type used to represent the different pretty printers
- the `printer` function that creates a value of the above type
- the `git_printer` value that prints textual diff similarly to `git diff`
- the `pp` function that takes a printer and returns a function to pretty print textual diffs

For now we'll simply re-implement those exactly like they are defined in `Diffutils`, we'll make adjustment to suit our needs in the future.

Those parts of the code are already tested so we won't need to write new ones there. Making sure our existing test pass should be enough.

azzsal:
This should close #79 

I have included types, functions and values from the diffutils dependency project into the Text_diffs module.

Thanks!

### DIFF
--- a/lib/text_diff.mli
+++ b/lib/text_diff.mli
@@ -1,11 +1,7 @@
 (** Utilities for custom diff printing  *)
 
-type conflict2 = { orig : string list; new_ : string list }
-(** In a value of type {!conflict2}, the sequence [orig] and [new_] should
-        have no common value. *)
-
-type hunk = Same of string | Diff of conflict2
-type t = hunk list
+type t
+(** Represents a single textual diff of a module interface*)
 
 type ts = t String_map.t
 (** Type for representing library interface diffs as text diff.
@@ -21,16 +17,8 @@ type ts = t String_map.t
     a new function [Main.M.do_something], this will show in the textual
     diff under the key ["Main.M"].
     Identical modules won't appear in the map.
+*)
 
-    Note that the individual [t] stored in the map only
-    contain line changes, i.e. only [Diff {orig; new_}] hunks, no [Same s]
-    ones. *)
-
-type printer
-
-val printer : same:string Fmt.t -> diff:conflict2 Fmt.t -> printer
-val git_printer : printer
-val pp : printer -> t Fmt.t
 val from_diff : Diff.module_ -> ts
 
 val pp_git : Format.formatter -> ts -> unit
--- a/lib/text_diff.mli
+++ b/lib/text_diff.mli
@@ -1,9 +1,8 @@
 (** Utilities for custom diff printing  *)
 
 type conflict2 = { orig : string list; new_ : string list }
-type diff = Diff of conflict2
 
-type t = diff list String_map.t
+type t = conflict2 list String_map.t
 (** Type for representing library interface diffs as text diff.
 
     Changes are arranged per fully qualified module path.
--- a/lib/text_diff.mli
+++ b/lib/text_diff.mli
@@ -1,13 +1,32 @@
 (** Utilities for custom diff printing  *)
 
-type t
-(** Represents a single textual diff of a module interface*)
+type conflict2 = { orig : string list; new_ : string list }
+type diff = Diff of conflict2
 
-val pp : Format.formatter -> Diff.module_ -> unit
+type t = diff list String_map.t
+(** Type for representing library interface diffs as text diff.
+
+    Changes are arranged per fully qualified module path.
+    Keys are module path, as strings, that map to the textual diff
+    for the content of said module.
+    
+    The removal or addition of a module is listed under its parent.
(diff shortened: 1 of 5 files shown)


panglesd:
Hello @azzsal !

Thanks for the contribution, that was quick!

It is already fixing #79, and so is mergeable, but since you were quick, I propose to make further improvements directly in this PR.

Indeed, the `Diffutils` library was made for representing arbitrary diffs. Here, we have a special case of diffs, so we now want to "tailor it to our special case". Here are different ways this can be improved:
- Try to make the `Text_diff` interface minimal. For this, you can look at the values that are used outside this module, and try to expose only those ones. Making a type abstract can also help removing many type definitions.
- Inside `Text_diif`, try to tailor the type definition to our usecase. E.g., remove any constructor/field that is not required in our case!

Any other idea is welcome. If possible, try to have atomic commits that each contain a single improvement!

And do not hesitate to ask any question if something is unclear!

azzsal:
Thank you so much for the detailed feedback!

I wasn't sure if I should make the interface minimal in the first place, It turns out I should've. I really appreciate the suggestion.
I believe I made the suggested improvements, but I am not sure whether I should keep the t map in the interface or not.

Thank you.

panglesd:
> I wasn't sure if I should make the interface minimal in the first place, It turns out I should've.

That was not required in the original issue! It's good to do as you did: start with few changes and improve over time. :+1:

> I am not sure whether I should keep the t map in the interface or not.

That's a good start, remove as much as possible, there are some things remaining :) I think you could even remove the `ts` type altogether, putting `from_diff` inside the `pp` values.

azzsal:
Hope this makes it as minimal as possible :)

lib/text_diff.mli
```diff
@@ -1,32 +1,13 @@
...
-    On the other hand, if [Main.M] is present in both versions but received
-    a new function [Main.M.do_something], this will show in the textual
-    diff under the key ["Main.M"].
-    Identical modules won't appear in the map.
-
-    Note that the individual [Diffutils.Diff.t] stored in the map only
-    contain line changes, i.e. only [Diff {orig; new_}] hunks, no [Same s]
-    ones. *)
-
-val from_diff : Diff.module_ -> Diffutils.Diff.t String_map.t
```
NathanReb, line 23: Could you please also re-expose this value and restore the orignal `pp` types so they take a `Text_diff.t`?

panglesd (approved):
Small nitpicks to resolve before merging!

EDIT: Just saw @NathanReb's review. @azzsal sorry for the confusion, should have aligned with Nathan beforehand!

lib/text_diff.mli
```diff
@@ -1,32 +1,13 @@
-(** Conversion from library diff abstraction to textual diff, line based
-    diff *)
+(** Utilities for custom diff printing  *)
 
-type t = Diffutils.Diff.t String_map.t
-(** Type for representing library interface diffs as text diff.
+type t
```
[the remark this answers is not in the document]
azzsal: Should I make the type t abstract as @panglesd suggested or include the type and the comment in the interface ?

lib/text_diff.mli
azzsal: Yes, I agree. It's better to separate the low-level diff from the textual one.

lib/text_diff.mli
```diff
@@ -1,7 +1,9 @@
-(** Conversion from library diff abstraction to textual diff, line based
-    diff *)
+(** Utilities for custom diff printing  *)
 
-type t = Diffutils.Diff.t String_map.t
+type conflict2 = { orig : string list; new_ : string list }
+type diff = Diff of conflict2
```
NathanReb, line 4: Since there is a single variant here, we can remove this type and turn our type `t` into:
```
type t = conflict2 list String_map.t
```

NathanReb (commented):
Sorry for the long delay! This is looking good. Indeed we have to expose our type `t` since we want external users to be able to manipulate it.

I made a small suggestion to simplify the set of types we use but other than that the PR looks ready to me!

NathanReb (approved):
Looks good, let's merge!

PR #89, opened 2024-10-07
Detect addition and removal of type declarations

### LINKED ISSUE #80: Detect removal and addition of type declarations

To get one step closer, we would like to add detection of the addition and removal of type_declaration items.

## How to proceed

To detect addition/removal of type declarations, we will have to extend the [`item_type`](https://github.com/ocaml-semver/ocaml-api-watch/blob/main/lib/diff.ml#L23) and [`sig_items`](https://github.com/ocaml-semver/ocaml-api-watch/blob/main/lib/diff.ml#L24) types which are used to build the map mentioned above. These are then used in the `extract_items` function that builds the map from the module signature.

## Testing

This should be tested at the library level. You can add a new `tests/api-watch/test_diff_type_decl.ml` file where you can write tests for type addition or removal. You can use the [`tests/api-watch/test_diff.ml`](https://github.com/ocaml-semver/ocaml-api-watch/blob/main/tests/api-watch/test_diff.ml) or [`tests/api-watch/test_diff_module.ml`](https://github.com/ocaml-semver/ocaml-api-watch/blob/main/tests/api-watch/test_diff_module.ml) files as examples on how to write such tests.

azzsal:
This PR aims to close #80. I will follow TDD. I have updated tests in  [tests/api-watch/test_diff.ml](https://github.com/ocaml-semver/ocaml-api-watch/blob/main/tests/api-watch/test_diff.ml) to detect the addition and removal type declarations.
The PR is still a work in progress and will updated accordingly.


azzsal:
I have updated the Diff module to include support for addition and removal of type declarations diffing. The tests cannot run since the TextDiff module need to be updated to handle type declarations. Before I go ahead and processed, could you review if my changes in the Diff module are good ?

Thank you!

lib/diff.ml
```diff
@@ -51,6 +62,15 @@ let modtype_item ~loc ~typing_env ~name ~reference ~current =
...
       Some (Module { mname = name; mdiff = Modified Unsupported })
 
+let type_item ~name ~reference ~current =
+  match (reference, current) with
+  | None, None -> None
+  | Some (Typ refType), None ->
+      Some (Type { tname = name; tdiff = Removed refType })
+  | None, Some (Typ curType) ->
+      Some (Type { tname = name; tdiff = Added curType })
+  | _ -> None
```
panglesd, line 72: You can make this last case more explicit, instead of using a wildcard. This makes it easier to read what case(s) remains, and lets OCaml's match exhaustiveness check tells you if you have forgotten to treat a case!

panglesd (commented):
Looks good to me to continue!

azzsal:
I have updated the rest of Text_diff module and tests to accommodate for type declarations. All tests now pass!

lib/text_diff.ml
```diff
@@ -36,6 +43,18 @@ let process_value_diff (val_diff : Diff.value) =
...
+  match type_diff.tdiff with
+  | Added td ->
+      [
+        Diffutils.Diff.Diff { orig = []; new_ = td_to_lines type_diff.tname td };
+      ]
+  | Removed td ->
+      [
+        Diffutils.Diff.Diff { orig = td_to_lines type_diff.tname td; new_ = [] };
+      ]
+  | Modified _ -> failwith "Unimplemented"
```
panglesd, line 56: Even though the detection of modified types is not implemented, the printing of modified type is simple enough to be implemented now!

lib/text_diff.ml
```diff
@@ -77,12 +96,19 @@ let from_diff (diff : Diff.module_) : Diffutils.Diff.t String_map.t =
...
+                  (function
+                    | None -> Some diff | Some existing -> Some (existing @ diff))
+                  acc'
             | Module sub_module_diff ->
                 let sub_module_path =
                   match sub_module_diff.mdiff with
                   | Modified _ -> module_path ^ "." ^ sub_module_diff.mname
                   | Added _ | Removed _ -> module_path
                 in
+
```
panglesd, line 111:
```suggestion
```

tests/api-diff/run.t
panglesd, line 1: Could you also add a test for upcoming modified types?

panglesd (approved):
Looks ready for merge, with minor comments below. 
You already need to rebase, and I'd prefer to merge #88 and rebase before merging this one.

We'll wait for @NathanReb's review as well.
In the meantime, I think you can work on detection of modified types.

tests/api-watch/test_diff_module.ml
```diff
@@ -85,7 +85,7 @@ let%expect_test "Modules with both supported and unsupported changes" =
   [%expect
     {|
     Some (Module Main: {Modified (Supported [ Value (x, Removed);
-    Module M: {Modified (Unsupported)}])})|}]
```
NathanReb, line 88: It might be good to change the code a bit here so that the test output still contains an unsupported change.

Once all of the items are handled this part of the test won't be useful anymore but for now, you should be able to simply replace the `type t` with an exception declaration:
```
type exn += Some_exn
```

None of the issues we opened will cover this so it should fix it for other similar PRs as well.

NathanReb (commented):
Looks great! I added one comment about one of the changes in the tests.

You might have to rebase on top of the latest changes though before fixing it as #86 has been merged now.

azzsal:
Sorry for the messy commits, but I am not familiar with rebase and more used to merge. The cram tests are not passing for some reason, I will try to fix that.

panglesd:
Indeed, your rebase seems to have gone wrong. There is now more conflicts and more commits!

(I'm personally more used to rebase that's why I suggested it, but you could have done a merge if you prefer!)

Now, my advice would be:
- Create a new branch starting on main,
- cherry-pick the commits you need from the old branch to this new branch, one by one, resolving conflicts as they come
- once everything seems fine, replace the old branch by the new one.

(But maybe there is a better way!)

azzsal:
This can now be safely closed, since all the work is moved to #92. Sorry about the confusion.

azzsal:
This can now be safely closed, since all the work is moved to https://github.com/ocaml-semver/ocaml-api-watch/pull/92. Sorry about the confusion.

PR #92, opened 2024-10-08
Detected addition and removal of type declarations

### LINKED ISSUE #83: Detect addition and removal of class type declarations

To get one step closer, we would like to add detection of the addition and removal of class_type_declaration items.

## How to proceed

## Testing

This should be tested at the library level. You can add a new `tests/api-watch/test_diff_class_type_decl.ml` file where you can write tests for type addition or removal. You can use the [`tests/api-watch/test_diff.ml`](https://github.com/ocaml-semver/ocaml-api-watch/blob/main/tests/api-watch/test_diff.ml) or [`tests/api-watch/test_diff_module.ml`](https://github.com/ocaml-semver/ocaml-api-watch/blob/main/tests/api-watch/test_diff_module.ml) files as examples on how to write such tests.

azzsal:
This should resolve #83 and extend it to include detection of modified type declarations, as @panglesd suggested.
This is a continuation on the work of #89. I have closed it, since the rebase gone wrong.

### DIFF
--- a/lib/diff.ml
+++ b/lib/diff.ml
@@ -30,7 +30,7 @@ type item_type = Value_item | Module_item | Type_item [@@deriving ord]
 type sig_items =
   | Val of value_description
   | Mod of module_declaration
-  | Typ of type_declaration
+  | Typ of type_declaration * Ident.t
 
 module Sig_item_map = Map.Make (struct
   type t = item_type * string [@@deriving ord]
@@ -41,11 +41,11 @@ let extract_items items =
     (fun tbl item ->
       match item with
       | Sig_module (id, _, mod_decl, _, _) ->
-          Sig_item_map.add (Module_item, Ident.name id) (Mod mod_decl, id) tbl
+          Sig_item_map.add (Module_item, Ident.name id) (Mod mod_decl) tbl
       | Sig_value (id, val_des, _) ->
-          Sig_item_map.add (Value_item, Ident.name id) (Val val_des, id) tbl
+          Sig_item_map.add (Value_item, Ident.name id) (Val val_des) tbl
       | Sig_type (id, type_decl, _, _) ->
-          Sig_item_map.add (Type_item, Ident.name id) (Typ type_decl, id) tbl
+          Sig_item_map.add (Type_item, Ident.name id) (Typ (type_decl, id)) tbl
       | _ -> tbl)
     Sig_item_map.empty items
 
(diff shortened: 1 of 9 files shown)


azzsal:
I have added the case for the detection of type declarations modification. The failing tests indicate that signature items are added and removed again, I am guessing that's due to the change of the map key type to Ident.t. Please have a look @panglesd.

panglesd:
> I am guessing that's due to the change of the map key type to item_type * Ident.t

Yes exactly, I think that is the issue! Two items of the same name are likely to get two different idents, in their respective compilation units.

azzsal:
I have returned the map key type to item_type * string, and stored the compilation unit id of an item in the map value.
All tests pass now, except for one test for a nested module signature.

panglesd:
This test is a test for dyslexia! The is a typo: `shpae` instead of `shape`.

azzsal:
Sorry, I should have noticed that! All tests now pass!

lib/diff.ml
```diff
@@ -12,16 +12,25 @@ type value = {
...
+}
+
 type module_ = {
   mname : string;
   mdiff : (module_declaration, module_modification) t;
 }
 
 and module_modification = Unsupported | Supported of sig_item list
-and sig_item = Value of value | Module of module_
+and sig_item = Value of value | Module of module_ | Type of type_
```
panglesd, line 26: Note: an alternate design would have been to add the `Ident` only in the `Type` constructor payload, instead of on any entry of the `Sig_item_map`. That would have prevented adding the ident on values and module, where they are not used, and limited the conflicts on the rebase of other PRs!

But, your solution is still ok.

panglesd (approved):
That is looking good, it is ready to be merged!

I'll merge very soon. In the meantime, on another branch based on this one, you can work on using `Includecore.report_type_mismatch` to display a bit more in what way the two types differ. (This won't always give super good results, but that is fine! We'll made our own reporter later)

lib/diff.ml
azzsal: Yes, that's a better design. but I think it should be more easily added to Typ constructor, and stored in ```Sig_item_map``` with Type_item item type only.

lib/diff.ml
```diff
@@ -51,6 +62,28 @@ let modtype_item ~loc ~typing_env ~name ~reference ~current =
...
+  match (reference, current) with
+  | None, None -> None
+  | Some (Typ (reference, _)), None ->
+      Some (Type { tname = name; tdiff = Removed reference })
+  | None, Some (Typ (current, _)) ->
+      Some (Type { tname = name; tdiff = Added current })
+  | Some (Typ (reference, refId)), Some (Typ (current, curId)) -> (
+      let type_coercion1 () =
+        Includecore.type_declarations ~equality:true ~loc:current.type_loc
+          typing_env ~mark:true name current (Pident curId) reference
```
NathanReb, line 75: Have you tried setting `mark` and `equality` to false? From what I gather they are just used to modify the typing env to set type usage and we have no use for this.

Could you try setting `mark` to false and not setting equality (it's an optional argument) to see if that still works?

NathanReb (commented):
This looks great! Thanks for all your work on this!

I have a single comment regarding some of the arguments of `Includecore.type_declarations` but other than, it's good to go!

Great work!

NathanReb:
Rebasing is a bit hard at first but it's in important skill as it helps keeping a clean history and is the prefered method in quite a few repositories.

It's also easier to do if you stick to logical and self contained commits. If each single commit is clearly named and has its intent well specified, builds and passes the tests it becomes clear what to do to re-apply it on top of modified codebase and whether you did it right in the end.

panglesd:
Can you add a changelog entry? For the format, you can see [this example](https://github.com/ocaml-semver/ocaml-api-watch/pull/95/files#diff-d975bf659606195d2165918f93e1cf680ef68ea3c9cab994f033705fea8238b2R11).

NathanReb (approved):
Awesome, thanks!

PR #93, opened 2024-10-09
Add support for module types

### LINKED ISSUE #81: Detect addition and removal of module types

To get one step closer, we would like to add detection of the addition and removal of modtype_declaration items.

## How to proceed

## Testing

This should be tested at the library level. You can add a new `tests/api-watch/test_diff_modtype_decl.ml` file where you can write tests for type addition or removal. You can use the [`tests/api-watch/test_diff.ml`](https://github.com/ocaml-semver/ocaml-api-watch/blob/main/tests/api-watch/test_diff.ml) or [`tests/api-watch/test_diff_module.ml`](https://github.com/ocaml-semver/ocaml-api-watch/blob/main/tests/api-watch/test_diff_module.ml) files as examples on how to write such tests.

NchamJosephMuam:
@NathanReb @panglesd 
This PR
closes #81

### DIFF
--- a/lib/diff.ml
+++ b/lib/diff.ml
@@ -19,18 +19,30 @@ type type_ = {
 
 type module_ = {
   mname : string;
-  mdiff : (module_declaration, module_modification) t;
+  mdiff : (module_declaration, signature_modification) t;
 }
 
-and module_modification = Unsupported | Supported of sig_item list
-and sig_item = Value of value | Module of module_ | Type of type_
+and modtype = {
+  mtname : string;
+  mtdiff : (modtype_declaration, signature_modification) t;
+}
 
-type item_type = Value_item | Module_item | Type_item [@@deriving ord]
+and signature_modification = Unsupported | Supported of sig_item list
+
+and sig_item =
+  | Value of value
+  | Module of module_
+  | Type of type_
+  | Modtype of modtype
+
+type item_type = Value_item | Module_item | Type_item | Modtype_item
+[@@deriving ord]
 
 type sig_items =
   | Val of value_description
   | Mod of module_declaration
   | Typ of type_declaration * Ident.t
+  | Modtype of modtype_declaration
 
 module Sig_item_map = Map.Make (struct
   type t = item_type * string [@@deriving ord]
@@ -42,6 +54,8 @@ let extract_items items =
       match item with
       | Sig_module (id, _, mod_decl, _, _) ->
           Sig_item_map.add (Module_item, Ident.name id) (Mod mod_decl) tbl
+      | Sig_modtype (id, mtd_decl, _) ->
+          Sig_item_map.add (Modtype_item, Ident.name id) (Modtype mtd_decl) tbl
       | Sig_value (id, val_des, _) ->
           Sig_item_map.add (Value_item, Ident.name id) (Val val_des) tbl
       | Sig_type (id, type_decl, _, _) ->
@@ -49,7 +63,7 @@ let extract_items items =
       | _ -> tbl)
     Sig_item_map.empty items
 
-let modtype_item ~loc ~typing_env ~name ~reference ~current =
+let module_type_fallback ~loc ~typing_env ~name ~reference ~current =
   let modtype_coercion1 () =
     Includemod.modtypes ~loc typing_env ~mark:Mark_both reference current
   in
@@ -121,7 +135,9 @@ let rec items ~reference ~current =
       | Module_item, reference, current ->
           module_item ~typing_env:env ~name ~reference ~current
       | Type_item, reference, current ->
-          type_item ~typing_env:env ~name ~reference ~current)
+          type_item ~typing_env:env ~name ~reference ~current
+      | Modtype_item, reference, current ->
+          module_type_item ~typing_env:env ~name ~reference ~current)
     ref_items curr_items
   |> Sig_item_map.bindings |> List.map snd
 
@@ -136,13 +152,38 @@ and module_item ~typing_env ~name ~reference ~current =
       module_declaration ~typing_env ~name ~reference ~current
   | _ -> assert false
 
+and module_type_item ~typing_env ~name ~reference ~current =
+  match (reference, current) with
+  | None, None -> None
+  | None, Some (Modtype curr_mtd) ->
+      Some (Modtype { mtname = name; mtdiff = Added curr_mtd })
+  | Some (Modtype ref_mtd), None ->
+      Some (Modtype { mtname = name; mtdiff = Removed ref_mtd })
+  | Some (Modtype ref_mtd), Some (Modtype curr_mtd) ->
+      modtype_declaration ~typing_env ~name ~reference:ref_mtd ~current:curr_mtd
+  | _ -> assert false
+
 and module_declaration ~typing_env ~name ~reference ~current =
-  match (reference.md_type, current.md_type) with
+  module_type ~typing_env ~name ~ref_module_type:reference.md_type
+    ~current_module_type:current.md_type ~reference_location:reference.md_loc
+
+and modtype_declaration ~typing_env ~name ~reference ~current =
+  match (reference.mtd_type, current.mtd_type) with
+  | Some ref_sub, Some curr_sub ->
+      module_type ~typing_env ~name ~ref_module_type:ref_sub
+        ~current_module_type:curr_sub ~reference_location:reference.mtd_loc
+  | Some _, None | None, Some _ ->
+      Some (Modtype { mtname = name; mtdiff = Modified Unsupported })
+  | None, None -> None
+
+and module_type ~typing_env ~name ~ref_module_type ~current_module_type
+    ~reference_location =
+  match (ref_module_type, current_module_type) with
   | Mty_signature ref_submod, Mty_signature curr_submod ->
       signatures ~typing_env ~reference:ref_submod ~current:curr_submod
       |> Option.map (fun mdiff -> Module { mname = name; mdiff })
   | ref_modtype, curr_modtype ->
-      modtype_item ~loc:reference.md_loc ~typing_env ~name
+      module_type_fallback ~loc:reference_location ~typing_env ~name
         ~reference:ref_modtype ~current:curr_modtype
 
 and signatures ~typing_env ~reference ~current =
@@ -162,5 +203,5 @@ and signatures ~typing_env ~reference ~current =
 
 let interface ~module_name ~reference ~current =
   let typing_env = Env.empty in
-  signatures ~typing_env ~reference ~current
-  |> Option.map (fun mdiff -> { mname = module_name; mdiff })
+  let sig_out = signatures ~typing_env ~reference ~current in
+  Option.map (fun mdiff -> { mname = module_name; mdiff }) sig_out
--- a/lib/diff.mli
+++ b/lib/diff.mli
@@ -14,11 +14,21 @@ type type_ = {
 
 type module_ = {
   mname : string;
-  mdiff : (Types.module_declaration, module_modification) t;
+  mdiff : (Types.module_declaration, signature_modification) t;
 }
 
-and module_modification = Unsupported | Supported of sig_item list
-and sig_item = Value of value | Module of module_ | Type of type_
+and modtype = {
+  mtname : string;
+  mtdiff : (Types.modtype_declaration, signature_modification) t;
+}
+
+and signature_modification = Unsupported | Supported of sig_item list
+
+and sig_item =
+  | Value of value
+  | Module of module_
+  | Type of type_
+  | Modtype of modtype
 
 val interface :
   module_name:string ->
--- a/lib/text_diff.ml
+++ b/lib/text_diff.ml
@@ -41,20 +41,20 @@ let md_to_lines name md =
   let module_str = "module " ^ name ^ ": " ^ Buffer.contents buf in
   CCString.lines module_str
 
-  let mtd_to_lines name mtd =
-    match mtd.mtd_type with
-    | Some m ->
-        let buf = Buffer.create 256 in
-        let formatter = Format.formatter_of_buffer buf in
-        Printtyp.modtype formatter m;
-        Format.pp_print_flush formatter ();
-        let module_type_str =
-          "module type " ^ name ^ " = " ^ Buffer.contents buf
-        in
-        CCString.lines module_type_str
-    | None ->
-        let abstract_module_type_str = "module type " ^ name in
-        CCString.lines abstract_module_type_str
+let mtd_to_lines name mtd =
+  match mtd.mtd_type with
+  | Some m ->
+      let buf = Buffer.create 256 in
+      let formatter = Format.formatter_of_buffer buf in
+      Printtyp.modtype formatter m;
+      Format.pp_print_flush formatter ();
+      let module_type_str =
+        "module type " ^ name ^ " = " ^ Buffer.contents buf
+      in
+      CCString.lines module_type_str
+  | None ->
+      let abstract_module_type_str = "module type " ^ name in
+      CCString.lines abstract_module_type_str
 
 let process_diff (diff : (_, _ Diff.atomic_modification) Diff.t) name to_lines =
   match diff with
--- a/lib/text_diff.ml
+++ b/lib/text_diff.ml
@@ -41,6 +41,21 @@ let md_to_lines name md =
   let module_str = "module " ^ name ^ ": " ^ Buffer.contents buf in
   CCString.lines module_str
 
+  let mtd_to_lines name mtd =
+    match mtd.mtd_type with
+    | Some m ->
+        let buf = Buffer.create 256 in
+        let formatter = Format.formatter_of_buffer buf in
+        Printtyp.modtype formatter m;
+        Format.pp_print_flush formatter ();
+        let module_type_str =
+          "module type " ^ name ^ " = " ^ Buffer.contents buf
+        in
+        CCString.lines module_type_str
+    | None ->
+        let abstract_module_type_str = "module type " ^ name in
+        CCString.lines abstract_module_type_str
+
 let process_diff (diff : (_, _ Diff.atomic_modification) Diff.t) name to_lines =
   match diff with
   | Added item -> [ { orig = []; new_ = to_lines name item } ]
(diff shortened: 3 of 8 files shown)


lib/diff.ml
```diff
@@ -110,6 +135,16 @@ and module_declaration ~typing_env ~name ~reference ~current =
       modtype_item ~loc:reference.md_loc ~typing_env ~name
         ~reference:ref_modtype ~current:curr_modtype
 
+and module_type_declaration ~typing_env ~name ~reference ~current =
```
panglesd, line 138: This function has some duplicated code with `module_declaration`, could you factor things to avoid this duplication?

lib/diff.ml
```diff
@@ -110,6 +135,16 @@ and module_declaration ~typing_env ~name ~reference ~current =
...
+  match (reference.mtd_type, current.mtd_type) with
+  | Some (Mty_signature ref_sub), Some (Mty_signature curr_sub) ->
+      signatures ~typing_env ~reference:ref_sub ~current:curr_sub
+      |> Option.map (fun mdiff -> Module { mname = name; mdiff })
+  | Some ref_sub, Some curr_sub ->
+      modtype_item ~loc:reference.mtd_loc ~typing_env ~name ~reference:ref_sub
+        ~current:curr_sub
+  | _ -> assert false
```
panglesd, line 146: The `mtd_type` field is an option because a module type can be abstract, just like a type:

```
(* in an interface file *)
module type S

module M : S
```

So the `None` cases are possible and you need to handle them!

lib/diff.ml
```diff
@@ -17,11 +17,20 @@ type module_ = {
...
+and sig_item = Value of value | Module of module_ | Modtype of modtype
+
+type item_type = Value_item | Module_item | Modtype_item [@@deriving ord]
 
-type item_type = Value_item | Module_item [@@deriving ord]
-type sig_items = Val of value_description | Mod of module_declaration
+type sig_items =
+  | Val of value_description
+  | Mod of module_declaration
+  | Mtdtype of modtype_declaration
```
panglesd, line 33: What does `Mtdtype` stand for? I would have expected something close to `Mty` or `ModType`  as name for the variant...

lib/text_diff.ml
panglesd, line 1: In this file also, try to factor things out as much as possible!

panglesd (commented):
This is looking great!

You've went further than what the issue was requiring, which was just additions and removal (not modifications). I left some comments in the code, the biggest things are:
- handle abstract module types
- factor code with modules.

NchamJosephMuam:
@panglesd  Thank you for the reviews.

lib/diff.ml
panglesd: > ```
> module type P = sig val x : int end  
>
> becomes 
>
> module type P
> ```

Yes, exactly. This is an API change, so you should return `Some`!

lib/diff.mli
```diff
@@ -1,22 +1,28 @@
+open Types
```
panglesd, line 1: I am not sure I like the global open, as we define here types of very similar name that what can be found in `Types`: for instance, `sig_item` comes from this file, but `signature` comes from `Types`. Similarly for `module`, `values`, etc.
It makes it harder to read, and raise the chance of an accidental shadowing.
Could you revert this open?

lib/text_diff.ml
```diff
@@ -34,6 +34,17 @@ let md_to_lines name md =
...
 
+let mtd_to_lines name mtd =
+  let buf = Buffer.create 256 in
+  let formatter = Format.formatter_of_buffer buf in
+  (match mtd.mtd_type with
+  | Some m ->
+      Printtyp.modtype formatter m;
+      Format.pp_print_flush formatter ()
+  | None -> ());
+  let module_type_str = "module type " ^ name ^ " = " ^ Buffer.contents buf in
```
panglesd, line 45: This prints `module type M =` for abstract module types!

lib/text_diff.ml
```diff
@@ -46,7 +57,49 @@ let process_value_diff (val_diff : Diff.value) =
...
+      List.fold_left
+        (fun acc' change ->
+          match (change : Diff.sig_item) with
+          | Value val_diff ->
+              let diff = process_value_diff val_diff in
+              String_map.update module_path
+                (function
+                  | None -> Some diff | Some existing -> Some (existing @ diff))
+                acc'
+          | Module sub_module_diff -> from_diff sub_module_diff
```
panglesd, line 91: It seems to me that you lose the correct path here!

panglesd (commented):
Small comments. I think you can still factor out `process_module_diff` and `process_module_type_diff`, fixing a small bug by doing so!

lib/diff.ml
```diff
@@ -17,11 +17,20 @@ type module_ = {
...
+and sig_item = Value of value | Module of module_ | Modtype of modtype
+
+type item_type = Value_item | Module_item | Modtype_item [@@deriving ord]
 
-type item_type = Value_item | Module_item [@@deriving ord]
-type sig_items = Val of value_description | Mod of module_declaration
+type sig_items =
+  | Val of value_description
+  | Mod of module_declaration
+  | ModType of modtype_declaration
```
NathanReb, line 33: We use snake case in OCaml, not camel case (weirdly enough!). Could you change that to `Modtype` or `Mod_type`. I'd like the former better as it matches what you used for the `item_type` variant.

lib/diff.ml
NathanReb: If that gets too complicated, you could also simply compare concrete module types and leave the handling of `abstract <-> concrete` diffs for a later PR. If you go down that road, I think it would still be good to add a test that shows that this change is not handled!

lib/diff.ml
```diff
@@ -101,13 +114,37 @@ and module_item ~typing_env ~name ~reference ~current =
...
+    ~current_module_type:current.md_type ~reference_location:reference.md_loc
+
+and module_type_declaration ~typing_env ~name ~reference ~current =
+  match (reference.mtd_type, current.mtd_type) with
+  | Some ref_sub, Some curr_sub ->
+      module_helper ~typing_env ~name ~ref_module_type:ref_sub
+        ~current_module_type:curr_sub ~reference_location:reference.mtd_loc
+  | _ -> assert false
+
+and module_helper ~typing_env ~name ~ref_module_type ~current_module_type
```
NathanReb, line 140: I think we can come up with a better name here. If we rename `modtype` to `module_type_fallback` and `module_helper` to `module_type` then it's both clearer what those functions are for and closer to the naming convention of the other functions in this module.

What do you think?

lib/diff.ml
```diff
@@ -101,13 +114,37 @@ and module_item ~typing_env ~name ~reference ~current =
...
+      module_type_declaration ~typing_env ~name ~reference:ref_mtd
+        ~current:curr_mtd
+  | _ -> assert false
+
 and module_declaration ~typing_env ~name ~reference ~current =
-  match (reference.md_type, current.md_type) with
+  module_helper ~typing_env ~name ~ref_module_type:reference.md_type
+    ~current_module_type:current.md_type ~reference_location:reference.md_loc
+
+and module_type_declaration ~typing_env ~name ~reference ~current =
```
NathanReb, line 133: I think that for the sake of clarity, we should stick to the compiler type names, even though those types have sometimes unclear naming convention.

We have a growing number of functions here so being careful with their names is becoming more important.

tests/api-diff/module_type_test.t
```diff
@@ -0,0 +1,122 @@
...
+  $ cat > orig_modtype.mli << EOF
+  > module type M = sig val x : float end
+  > type ('a, 'b) result = Ok of 'a | Error of 'b
+  >  val a : string -> int
+  > val f : int -> string
+  > module type D = sig
+  >   val b : int list -> int
+  >   val g : int -> string
+  > end
+  > module P : sig val a : int end
```
NathanReb, line 84: There is a lot of noise in this test, could you simplify it to simply contain one module type that's modified across the two versions?

tests/api-watch/test_diff_modtype_decl.ml
```diff
@@ -0,0 +1,98 @@
+open Api_watch
+open Test_helpers
+
+let%expect_test "Module types with both value and submodule changes" =
```
NathanReb, line 4: This title is not very clear, what do you mean by that? What type of change are you testing?

tests/api-watch/test_diff_modtype_decl.ml
```diff
@@ -0,0 +1,98 @@
...
+  let reference =
+    compile_interface
+      {|
+    type ('a, 'b) result = Ok of 'a | Error of 'b
+    val f : int -> string
```
NathanReb, line 9: Same remark as with the cram tests. This can be cleared out as it is not required to test that we correctly diff modtypes.

Same goes to the other tests below. Having minimal tests is important to clearly exhibit the behaviour under test and to prevent test breakage from happening too often. For instance this test would break if we changed how we represent or display value changes even though that's not what we're testing here.

NathanReb (commented):
This looks great! There are still a few things to improve but this is clearly on the right tracks!

A lot of my comments revolve around the tests that are a bit cluttered and would benefit from being simpler and more focused on module types. There are also missing simple tests like an identity one (e.g. ensuring that two equal module types produce no diffs).

lib/diff.ml
```diff
@@ -43,13 +43,13 @@ let extract_items items =
...
       | Sig_modtype (id, mtd_decl, _) ->
-          Sig_item_map.add (Modtype_item, Ident.name id) (ModType mtd_decl) tbl
+          Sig_item_map.add (Modtype_item, Ident.name id) (Modtype mtd_decl) tbl
       | Sig_value (id, val_des, _) ->
           Sig_item_map.add (Value_item, Ident.name id) (Val val_des) tbl
       | _ -> tbl)
     Sig_item_map.empty items
 
-let modtype_item ~loc ~typing_env ~name ~reference ~current =
+let moddule_type_fallback ~loc ~typing_env ~name ~reference ~current =
```
panglesd, line 52: Those `d`s seem dduplicated!

lib/diff.ml
```diff
@@ -101,13 +114,50 @@ and module_item ~typing_env ~name ~reference ~current =
...
-  match (reference.md_type, current.md_type) with
+  module_type ~typing_env ~name ~ref_module_type:reference.md_type
+    ~current_module_type:current.md_type ~reference_location:reference.md_loc
+
+and modtype_declaration ~typing_env ~name ~reference ~current =
+  match (reference.mtd_type, current.mtd_type) with
+  | Some ref_sub, Some curr_sub ->
+      module_type ~typing_env ~name ~ref_module_type:ref_sub
+        ~current_module_type:curr_sub ~reference_location:reference.mtd_loc
+  | Some _ref_sub, None ->
```
NchamJosephMuam, line 137: @panglesd 
Is this moving in the right direction?

panglesd: It is hard to say where it is moving, but you are close to the solution I'd be happy in this PR! Which is a solution which does not modify the types that you have defined.

Recall that `Some ...` means that the module type is not abstract: `module type S = sig ... end`.
`None` means that the module type is abstract: `module type S`.

So, when we are in the case where `current, reference` is `None, Some _` it means that the we went from an abstract module type to one with a signature. Does that best correspond to an addition removal or modification of a module type?
If it is an addition, what is added?
If it is a removal, what is removed?
If it is a modification, you have two ways to describe the modification: either by giving the list of modification (the `Supported` constructor, or by saying that the change is currently unsupported (the `Unsupported` constructor. Which is applicable here?

Try to think "high level", and only do the implementation when you know what you want to represent. Note that it is perfectly fine to say that a modification is `Unsupported`, supporting new kinds of modification is definitely for another PR!

panglesd (commented):
Looking quite good! The factorization in `diff` is exactly what I was looking for!

Two things:
- The `process_module_type_diff` now seems to handle the path well. However, it can still be factored out with `process_module_diff` to avoid duplication of code.
- You will need to rebase your work on top of `main` at some point!

NchamJosephMuam:
Hi @panglesd @NathanReb 
I tried to make some updates based on the reviews you left. Sorry it takes some time. I am still learning OCaml and I am happy to be learning more during this contribution phase. 

Is there something I can add to this PR to make it better?
I've tried to refactor the code as much as possible to reduce duplication.
I've also filtered out the test to remove a lot of noise and just test specific things
I've tried to represent diff changes for all the common scenarios (I don't know if this is going in the direction you envision)

lib/text_diff.ml
```diff
@@ -41,6 +41,23 @@ let md_to_lines name md =
...
+  | Some m ->
+      Printtyp.modtype formatter m;
+      Format.pp_print_flush formatter ();
+      let module_type_str =
+        "module type " ^ name ^ " = " ^ Buffer.contents buf
+      in
+      CCString.lines module_type_str
+  | None ->
+      Printtyp.modtype_declaration (Ident.create_local name) formatter mtd;
+      Format.pp_print_flush formatter ();
```
NathanReb, line 57: This is unused in this branch, you can remove those lines!

NathanReb: You can also move the buffer and formatter bindings to the `| Some m ->` branch since they are not used here either!

tests/api-diff/module_type_test.t
```diff
@@ -0,0 +1,136 @@
...
+  diff module Modified_modtype:
+  -val a : string -> int
+  +val a : string -> float
+  
+  diff module Modified_modtype.M:
+  -val x : float
+  +val x : int
+  
+  [1]
+
```
NathanReb, line 103: It's probably worth adding another section here:
```suggestion

# Switching a module type from concrete to abstract

```

NathanReb (commented):
This is starting to look good!

I think the unit tests should be a bit more aligned with the cram tests in terms of the diffing code branches they are testing.

I think you should rebase before going any further though. You can read some of the instructions I wrote [here](https://github.com/ocaml-semver/ocaml-api-watch/pull/90#issuecomment-2410350062) if you're not familiar with the process.

tests/api-diff/module_type_test.t
```diff
@@ -0,0 +1,136 @@
...
+  $ api-diff orig_modtype.cmi modified_modtype.cmi
+  diff module Modified_modtype:
+  -val a : string -> int
+  +val a : string -> float
+  
+  diff module Modified_modtype.M:
+  -val x : float
+  +val x : int
+  
+  [1]
```
NathanReb, line 102: You can remove this test as it is not testing anything new compared to the other tests above 
```suggestion
```

NchamJosephMuam:
@NathanReb @panglesd Is there something I can add  to the PR to get it ready for a merge?

lib/diff.mli
```diff
@@ -17,8 +17,18 @@ type module_ = {
   mdiff : (Types.module_declaration, module_modification) t;
 }
 
+and modtype = {
+  mtname : string;
+  mtdiff : (Types.modtype_declaration, module_modification) t;
+}
+
 and module_modification = Unsupported | Supported of sig_item list
```
panglesd, line 25: I think that this should have been named `signature_modification` and now that module types uses it, it is more visible.

Can you rename it to `signature` ?

(Both a module type and a module have a signature, which is a list of signature items)

lib/text_diff.ml
```diff
@@ -65,54 +80,86 @@ let process_type_diff (type_diff : Diff.type_) =
...
+        acc
+  | Removed ref_md ->
+      let diff =
+        [ { orig = md_to_lines module_diff.mname ref_md; new_ = [] } ]
+      in
+      String_map.update module_path
+        (function None -> Some diff | Some existing -> Some (existing @ diff))
+        acc
+  | Modified (Supported changes) ->
+      module_modification_changes module_path changes acc
```
panglesd, line 128: Those two functions (`process_module_type_diff` and `process_module_diff`) are still very similar.

Factoring them would be make the code much better I think. However, this could be done in another PR, not to delay this one.

If you want to try to factor these function, what you can do is:
- extract all the common part into a function
- fill the non-common part with variables, passed as input
- call the function from `process_module_diff` and `process_module_type_diff` with the right arguments
- Report any problem you (will) have ;)

lib/text_diff.ml
```diff
@@ -65,54 +80,86 @@ let process_type_diff (type_diff : Diff.type_) =
...
+
+and module_modification_changes module_path items acc =
```
panglesd, line 130: This should be `signature_change`, following my previous comment on signature!

panglesd (approved):
Looks great!

I've left:
- A name change to better highlight that something applies to signature rather than to modules
- A factoring proposition. You can decide to do it in this PR or in another (which could be done by someone else if you want), all solutions are fine!

Once you do the name change and decide what solution you prefer, this PR is ready to be merged: thanks!

lib/text_diff.ml
NchamJosephMuam: I've tried to refactor the code to avoid duplication. Is it in a better state now? Is there something I can add ?

panglesd: That is not exactly what I had in mind, but it is fine. Basically, you factored out everything, except the function `process_module_diff` and `process_module_type_diff` :wink:.

Here is what I had in mind:

```ocaml
let rec process_sig_diff :
    type a. _ -> (string -> a -> string list) -> (a, _) Diff.t * _ -> _ -> _ =
 fun path to_lines ((diff : (a, _) Diff.t), name) acc ->
  match diff with
  | Added curr_mtd ->
      let diff = [ { orig = []; new_ = to_lines name curr_mtd } ] in
      String_map.update path
        (function None -> Some diff | Some existing -> Some (existing @ diff))
        acc
  | Removed ref_mtd ->
      let diff = [ { orig = to_lines name ref_mtd; new_ = [] } ] in
      String_map.update path
        (function None -> Some diff | Some existing -> Some (existing @ diff))
        acc
  | Modified Diff.Unsupported ->
      String_map.add path
        [ { orig = []; new_ = [ "<unsupported change>" ] } ]
        acc
  | Modified (Supported changes) -> module_modification_changes path changes acc

and process_module_type_diff module_path (module_type_diff : Diff.modtype) acc =
  process_sig_diff module_path mtd_to_lines
    (module_type_diff.mtdiff, module_type_diff.mtname)
    acc

and process_module_diff module_path (module_diff : Diff.module_) acc =
  process_sig_diff module_path md_to_lines
    (module_diff.mdiff, module_diff.mname)
    acc

and module_modification_changes module_path items acc =
 [...]
```

the `type a . <type>` is needed because I call the function `process_sig_diff` recursively with different types. See [here](https://ocaml.org/manual/5.2/gadts-tutorial.html#s:gadts-recfun) for more information!

Your factoring of the printing function is good, but we should keep one printing function per type (`md_to_lines`, `mtd_to_lines`, ...) to avoid dealing with printing in the middle of some other logic (choosing with delimiter to use, ...).

panglesd: (That would be even more unrelated to this PR, but one could factor the `process_value_diff` and `process_type_diff` and divide by two the number of lines they take by removing the duplicated logic.)

NchamJosephMuam: I tried to do this but kept having a tpye mismatch. Are you using a gadts? that's awesome

NchamJosephMuam: Thank you. I just saw the link to the documentation you attached.

panglesd (approved):
Looks perfect! No more nitpick on my side.

The light is green on my side.

NchamJosephMuam:
@panglesd should i do this refactoring for value and type diff in a new Pr or add it here:

```ocaml
let process_diff :
    type a.
    (string -> a -> string list) ->
    (a, a Diff.atomic_modification) Diff.t ->
    _ ->
    _ =
 fun to_lines (diff : (a, a Diff.atomic_modification) Diff.t) name ->
  match diff with
  | Added item -> [ { orig = []; new_ = to_lines name item } ]
  | Removed item -> [ { orig = to_lines name item; new_ = [] } ]
  | Modified { reference; current } ->
      [ { orig = to_lines name reference; new_ = to_lines name current } ]

let process_value_diff (val_diff : Diff.value) =
  process_diff vd_to_lines val_diff.vdiff val_diff.vname

let process_type_diff (type_diff : Diff.type_) =
  process_diff td_to_lines type_diff.tdiff type_diff.tname
```

tests/api-watch/test_diff_modtpe_decl.ml
```diff
@@ -0,0 +1,75 @@
+open Api_watch
+open Test_helpers
+
+let%expect_test "Module_type with value changes" =
```
NathanReb, line 4: The test name does not correspond to what we are testing here. If you want to test for the removal of a value, note that you can define an empty module type:
```ocaml
module type T = sig
end
```

If you wish to test a module type going from concrete to abstract, then the test name should reflect that.

tests/api-watch/test_diff_modtpe_decl.ml
```diff
@@ -0,0 +1,75 @@
...
+  in
+  let current = compile_interface {|
+    module type M 
+  |} in
+  let result = Diff.interface ~module_name:"Main" ~reference ~current in
+  Format.printf "%a" pp_diff_option result;
+  [%expect
+    {| Some (Module Main: {Modified (Supported [ Module_type M: {Modified (Unsupported)}])}) |}]
+
+let%expect_test "Module_type with module change" =
```
NathanReb, line 21: What do you mean by `module change` here? I think this could be simplified to `"Modified module type"`.

tests/api-watch/test_diff_modtpe_decl.ml
```diff
@@ -0,0 +1,75 @@
...
+  let reference =
+    compile_interface
+      {|
+    module type M = sig
+      val b : int list -> int
+    end
+    module P : sig val x: float end
```
NathanReb, line 28: You can remove this as it is just noise for the test. It polutes the output with a removed module which is already tested elsewhere.

tests/api-watch/test_diff_modtpe_decl.ml
```diff
@@ -0,0 +1,75 @@
...
+  module type P = sig val x : int end
+  |}
+  in
+  let current = compile_interface {|
+  module type P
+  |} in
+  let result = Diff.interface ~module_name:"Main" ~reference ~current in
+  Format.printf "%a" pp_diff_option result;
+  [%expect
+    {| Some (Module Main: {Modified (Supported [ Module_type P: {Modified (Unsupported)}])}) |}]
```
NathanReb, line 61: One of this one or the first test can be removed as they are just duplicates.

tests/api-watch/test_diff_modtpe_decl.ml
```diff
@@ -0,0 +1,75 @@
...
+    module P : sig val x: float end
+  |}
+  in
+  let current =
+    compile_interface
+      {|
+    module type M = sig
+      val b : float list -> float
+    end
+    module type N = sig val x: int end
```
NathanReb, line 37: Did you add this module type to test module type addition? It's best to keep test cases minimal and focused on a specific feature, code path or regression.

Module type addition should have its own test case by comparing an empty interface with one that has a module type.

NathanReb (commented):
It looks good! I just have a few comments regarding the unit tests.

There are no tests for the module type removal detection. It's indirectly tested by another test in `test_diff_module.ml` but it would be good to have a dedicated test in `test_diff_module_type.ml` as the former could be changed.

tests/api-watch/test_diff_modtpe_decl.ml
NchamJosephMuam: Thank you

tests/api-watch/test_diff_modtpe_decl.ml
```diff
@@ -0,0 +1,68 @@
...
+      val b : float list -> float
+    end
+  |}
+  in
+  let result = Diff.interface ~module_name:"Main" ~reference ~current in
+  Format.printf "%a" pp_diff_option result;
+  [%expect
+    {| Some (Module Main: {Modified (Supported [ Module M: {Modified (Supported [ Value (b, Modified)])}])}) |}]
+
+let%expect_test "Concrete Module_type to abstract module_type change" =
```
NathanReb, line 43:
```suggestion
let%expect_test "Concrete to abstract module_type" =
```

tests/api-watch/test_diff_modtpe_decl.ml
```diff
@@ -0,0 +1,68 @@
...
+    end
+  |}
+  in
+  let current = compile_interface {|
+    module type M = sig end
+  |} in
+  let result = Diff.interface ~module_name:"Main" ~reference ~current in
+  Format.printf "%a" pp_diff_option result;
+  [%expect
+    {| Some (Module Main: {Modified (Supported [ Module M: {Modified (Supported [ Value (g, Removed)])}])}) |}]
```
NathanReb, line 19: The first two tests are testing the same thing, i.e. a module type whose signature has been modified. You can delete this one.

NathanReb (commented):
Can you please add a test case for module type removal?

NathanReb (approved):
Looks good, thanks! Let's merge :tada:

PR #98, opened 2024-10-13
Abstract away Sig_item_map module

### LINKED ISSUE #96: Abstract away Sig_item_map
Our Sig_item_map module is growing and its current implementation and API are not the best as they allow incorrect combination of keys and bindings.

We would like to extract it to its own module and improve its API so that the user (us) does not have to deal with those impossible cases.

To keep things simple we'll likely need a GADT (you can read more about them in the [OCaml manual](https://ocaml.org/manual/5.2/gadts-tutorial.html) or [RealWorldOCaml](https://dev.realworldocaml.org/gadts.html)).

I propose the following API:
```ocaml
type t

type _ item_type =
  | Value : value_declaration item_type
  | Module : module_declaration item_type
  ...

val empty : t

val add : name: string -> 'a item_type -> 'a -> t -> t

val diff : diff_item: ('a item_type -> 'a option -> 'a option -> 'diff option) -> t -> t -> 'diff list
```

The simplest and most straightforward implementation would be to build one internal map per item_type as suggested by @panglesd.

azzsal:
This should eventually resolve #96. I have implemented the empty value and the add function. The tests are failing since the diff function is yet to be implemented. @panglesd @NathanReb Please check if I am on the right track.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,8 +8,6 @@
 
 ### Changed
 
-- Abstracted Sig_item_map into is own module (#98, @azzsal)
-
 ### Deprecated
 
 ### Fixed
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,8 @@
 
 ### Changed
 
+- Abstracted Sig_item_map into is own module (#98, @azzsal)
+
 ### Deprecated
 
 ### Fixed
--- a/lib/diff.ml
+++ b/lib/diff.ml
@@ -75,11 +75,11 @@ let module_type_fallback ~loc ~typing_env ~name ~reference ~current =
 let type_item ~typing_env ~name ~reference ~current =
   match (reference, current) with
   | None, None -> None
-  | Some (Typ (reference, _)), None ->
+  | Some (reference, _), None ->
       Some (Type { tname = name; tdiff = Removed reference })
-  | None, Some (Typ (current, _)) ->
+  | None, Some (current, _) ->
       Some (Type { tname = name; tdiff = Added current })
-  | Some (Typ (reference, refId)), Some (Typ (current, curId)) -> (
+  | Some (reference, refId), Some (current, curId) -> (
       let type_coercion1 () =
         Includecore.type_declarations ~loc:current.type_loc typing_env
           ~mark:false name current (Pident curId) reference
@@ -92,16 +92,14 @@ let type_item ~typing_env ~name ~reference ~current =
       | None, None -> None
       | _, _ ->
           Some (Type { tname = name; tdiff = Modified { reference; current } }))
-  | _ -> None
 
 let value_item ~typing_env ~name ~reference ~current =
   match (reference, current) with
   | None, None -> None
-  | Some (Val reference), None ->
+  | Some reference, None ->
       Some (Value { vname = name; vdiff = Removed reference })
-  | None, Some (Val current) ->
-      Some (Value { vname = name; vdiff = Added current })
-  | Some (Val reference), Some (Val current) -> (
+  | None, Some current -> Some (Value { vname = name; vdiff = Added current })
+  | Some reference, Some current -> (
       let val_coercion1 () =
         Includecore.value_descriptions ~loc:current.val_loc typing_env name
           current reference
@@ -117,35 +115,30 @@ let value_item ~typing_env ~name ~reference ~current =
       | exception Includecore.Dont_match _ ->
           Some (Value { vname = name; vdiff = Modified { reference; current } })
       )
-  | _ -> None
 
 let rec items ~reference ~current =
   let env = Typing_env.for_diff ~reference ~current in
   let ref_items = extract_items reference in
   let curr_items = extract_items current in
-  (* Sig_item_map.merge
-       (fun (item_type, name) ref_opt curr_opt ->
-         match (item_type, ref_opt, curr_opt) with
-         | Value_item, reference, current ->
-             value_item ~typing_env:env ~name ~reference ~current
-         | Module_item, reference, current ->
-             module_item ~typing_env:env ~name ~reference ~current
-         | Type_item, reference, current ->
-             type_item ~typing_env:env ~name ~reference ~current
-         | Modtype_item, reference, current ->
-             module_type_item ~typing_env:env ~name ~reference ~current)
-       ref_items curr_items
-     |> Sig_item_map.bindings |> List.map snd *)
-  []
+  Sig_item_map.diff
+    ~diff_item:(fun item_type name reference current ->
+      match item_type with
+      | Sig_item_map.Value ->
+          value_item ~typing_env:env ~name ~reference ~current
+      | Sig_item_map.Module ->
+          module_item ~typing_env:env ~name ~reference ~current
+      | Sig_item_map.Modtype ->
+          module_type_item ~typing_env:env ~name ~reference ~current
+      | Sig_item_map.Type -> type_item ~typing_env:env ~name ~reference ~current)
+    ref_items curr_items
 
-and module_item ~typing_env ~name ~reference ~current =
+and module_item ~typing_env ~name ~(reference : module_declaration option)
+    ~(current : module_declaration option) =
   match (reference, current) with
   | None, None -> None
-  | None, Some (Mod curr_md) ->
-      Some (Module { mname = name; mdiff = Added curr_md })
-  | Some (Mod ref_md), None ->
-      Some (Module { mname = name; mdiff = Removed ref_md })
-  | Some (Mod reference), Some (Mod current) ->
+  | None, Some curr_md -> Some (Module { mname = name; mdiff = Added curr_md })
+  | Some ref_md, None -> Some (Module { mname = name; mdiff = Removed ref_md })
+  | Some reference, Some current ->
       module_declaration ~typing_env ~name ~reference ~current
   | _ -> assert false
 
--- a/lib/diff.ml
+++ b/lib/diff.ml
@@ -117,12 +117,10 @@ let rec items ~reference ~current =
   let diff_item : type a. (a, 'diff) Sig_item_map.diff_item =
    fun item_type name reference current ->
     match item_type with
-    | Sig_item_map.Value -> value_item ~typing_env:env ~name ~reference ~current
-    | Sig_item_map.Module ->
-        module_item ~typing_env:env ~name ~reference ~current
-    | Sig_item_map.Modtype ->
-        module_type_item ~typing_env:env ~name ~reference ~current
-    | Sig_item_map.Type -> type_item ~typing_env:env ~name ~reference ~current
+    | Value -> value_item ~typing_env:env ~name ~reference ~current
+    | Module -> module_item ~typing_env:env ~name ~reference ~current
+    | Modtype -> module_type_item ~typing_env:env ~name ~reference ~current
+    | Type -> type_item ~typing_env:env ~name ~reference ~current
   in
   Sig_item_map.diff ~diff_item:{ diff_item } ref_items curr_items
 
--- a/lib/diff.ml
+++ b/lib/diff.ml
@@ -114,23 +114,17 @@ let rec items ~reference ~current =
   let env = Typing_env.for_diff ~reference ~current in
   let ref_items = extract_items reference in
   let curr_items = extract_items current in
-  Sig_item_map.diff
-    ~diff_item:
-      {
-        diff_item =
-          (fun (type a) (item_type : a Sig_item_map.item_type) name
-               (reference : a option) (current : a option) ->
-            match item_type with
-            | Sig_item_map.Value ->
-                value_item ~typing_env:env ~name ~reference ~current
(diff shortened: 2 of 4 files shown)


lib/sig_item_map.ml
```diff
@@ -0,0 +1,25 @@
+open Types
+
+type t =
+  value_description String_map.t
+  * module_declaration String_map.t
+  * (type_declaration * Ident.t) String_map.t
```
NathanReb, line 6: You should use a record here instead of a tuple. The types being different for each of the members does not allow mistakes but a record will still make the code easier to read!

lib/sig_item_map.ml
```diff
@@ -0,0 +1,25 @@
...
+  | Value : value_description item_type
+  | Module : module_declaration item_type
+  | Type : (type_declaration * Ident.t) item_type
+
+let empty : t = (String_map.empty, String_map.empty, String_map.empty)
+
+let add : type a. name:string -> a item_type -> a -> t -> t =
+ fun ~name item_type item maps ->
+  match (item_type, maps) with
+  | Value, (values_map, modules_map, types_map) ->
```
NathanReb, line 18:
```suggestion
  | Value, (values_map, _, _) ->
```
or once you made it a record:
```suggestion
  | Value, {values_map; _} ->
```

lib/diff.ml
```diff
@@ -113,17 +116,21 @@ let rec items ~reference ~current =
...
-          value_item ~typing_env:env ~name ~reference ~current
-      | Module_item, reference, current ->
-          module_item ~typing_env:env ~name ~reference ~current
-      | Type_item, reference, current ->
-          type_item ~typing_env:env ~name ~reference ~current)
-    ref_items curr_items
-  |> Sig_item_map.bindings |> List.map snd
+  let values_diff = [] in
+  let modules_diff = [] in
+  let types_diff = [] in
```
NathanReb, line 121: You won't need 3 distinct calls here, just one that will return all diffs. The fact that we have different maps is an implementation detail, from the outside we just access a single data structure holding all type of items.

NathanReb (commented):
You seem to be on the right tracks indeed, well done! I made a couple small suggestions!

azzsal:
@NathanReb Thank for the suggestions. I have taken them into account.
I have added an initial implementation of the diff function, but the there is an error I couldn't understand.
error
The first error suggests that ```a type_item``` is a concrete type, not an abstract type, the outer ```type a``` seems to be unrecognized as an abstract type in the diff_item function type. Please point me to resources about the issue.

panglesd:
First, here is small list of resources on GADTs, which I'm sure you have already read.

 - The language manual : https://ocaml.org/manual/5.2/gadts-tutorial.html
 - The real world of ocaml chapter on GADTs : https://dev.realworldocaml.org/gadts.html
 - Another tutorial on GADTs : https://discuss.ocaml.org/t/tutorial-on-gadts/14040

Now, here are some resources specific to your problem: 
- https://ocaml.org/manual/5.2/polymorphism.html#s%3Ahigher-rank-poly
- https://discuss.ocaml.org/t/type-of-explicitly-polymorphic-function/11604/3

GADTs are notoriously hard! And those resources on your specific problem are a bit hidden when you don't know exactly what to search. Let me know if you have further problem!
(Your current implementation is already very good and I appreciate that you asked for resources instead of a solution!)

NathanReb:
The section you'll be interested in is this one: https://ocaml.org/manual/5.2/gadts-tutorial.html#s:gadts-recfun

The `.` in the type annotation is important as that's the part that means "For all types a".

NathanReb:
Ok, my suggestion is a bit misleading. To better understand the situation here I suggest you read the manual chapter on polymoprhism: https://ocaml.org/manual/5.2/polymorphism.html and in particular this section: https://ocaml.org/manual/5.2/polymorphism.html#s:higher-rank-poly.

panglesd:
Let me just emphasize that your problem actually has nothing to do with GADTs (although, this problem comes often in the case of a GADT.

When typing a structure item, OCaml introduces type variables. For instance, given:

```
let apply_and_add f x y = f x + f y
```

the typer starts by introducing type variables for `f`, `x`, `y`, for instance `'a`, `'b` and `'c`. Then, it "infer" constraints on the types by reasoning on the expression. For instance, the fact that `f` is applied with `x` says that actually, `'a` must be of the form `'b -> 'd`. Similarly, `'a` must be of the form `'c -> 'e`. Therefore, `'b = 'c` and `'d = 'e`. The `+` says that `'d = int`.

In the end, the type for `apply_and_add` is `('a -> int) -> 'a -> 'a -> int`. There is an unbound variable here: `'a`. The typer "quantify" it by interpreting the type as "For every type `'a`, `apply_and_add` has the type `('a -> int) -> 'a -> 'a -> int`". This quantification is written with a `.` in OCaml: in our example, `'a . ('a -> int) -> 'a -> 'a -> int`  This allows you to apply it in various context:

```ocaml
let square x = x * x
let _ = apply_and_add square 3 4
(* Works, because the quantified ['a] can be instanciated with type [int].
   Result is is 25 *)
let _ = apply_and_add List.length [1;2;3] [1;2]
(* Similar. Result is 5 *)
let _ = apply_and_add List.length [1;2;3] ["1";"2"]
(* Does not work: ['a] cannot be instanciated to any type that would make that work. *)
```

However, this quantification happens at the end of the typing of a structure item. In particular, the typer will nevel come up by itself with a quantification in the middle of a type, such as: `('x. 'x -> int) -> 'a -> 'b -> int`. Having this would have made the previous example work...

In your case, you are applying `diff_item` once with `Value` (which has some type) and once with `Module` (which has a _different_ type), so the typer fails to type the expression (since the typer cannot introduce a "forall" in the type of `diff_item`).

More information on the resources [I](https://ocaml.org/manual/5.2/polymorphism.html#s%3Ahigher-rank-poly) [linked](https://discuss.ocaml.org/t/type-of-explicitly-polymorphic-function/11604/3)!

panglesd:
> The section you'll be interested in is this one: https://ocaml.org/manual/5.2/gadts-tutorial.html#s:gadts-recfun

(To read only if you are confused with the syntax used in the link above! You do not need to know this to fix your error.)

One last precision on syntax, since it can be confusing.

You already know the locally abstract type syntax: `(type a)`. I just introduced you the "forall" syntax (for instance `'a . 'a -> 'a`).
In the context of recursive function working with GADTs (not your case), there is a third syntax, which is actually just a combination of the two syntax you know: the `type a . ...` syntax.

Quoting this [post](https://discuss.ocaml.org/t/how-to-read-understand-function-signature-with-gadt/2250/3):

```
let rec eval: type t. t term -> t = function
```
is actually a shorthand for the exactly equivalent:
```
let rec eval: 't. 't term -> 't = 
fun (type t) (x: t term): t ->
```

(But again, you don't need that!)

azzsal:
@panglesd @NathanReb Thank you so much for the detailed explanations, they clarified a lot of things for me. I'll take the time to go through the linked resources to better understand parametric polymorphism in OCaml. I hope as I am working on the project to get a deeper understanding of the type checking and type inference.
I think I have found a solution to the problem from quickly reading [Higher-rank polymorphic funcitons](https://ocaml.org/manual/5.2/polymorphism.html#s%3Ahigher-rank-poly). I update the code to reflect that. Thanks again!

lib/sig_item_map.ml
```diff
@@ -40,8 +40,12 @@ let add (type a) ~name (item_type : a item_type) (item : a)
         types_map = String_map.add name item types_map;
       }
 
-let diff (type a)
-    ~(diff_item : a item_type -> string -> a option -> a option -> 'diff option)
+type 'diff poly_diff_item = {
+  diff_item :
+    'a. 'a item_type -> string -> 'a option -> 'a option -> 'diff option;
```
panglesd, line 45: This type is long and written in several places.

Let's expose it as a (parametrized) type alias, to avoid duplication and make things more readable!

panglesd (commented):
Great!

I've left a small comment on style, but this looks great.

(I'll do a more thorough review at some point, but only for nitpicks. Also, we might want to wait a bit for some other PRs to be merged first, since it might be easier to rebase/merge this PR on them than the contrary)

lib/sig_item_map.ml
panglesd: What I mean is to define `type ('a, 'diff) diff_item = 'a item_type -> string -> 'a option -> 'a option -> 'diff option` first, expose it in the interface, and use it anywhere needed (I did not give much though to the name).
In particular, this can simplify [these line](https://github.com/ocaml-semver/ocaml-api-watch/pull/98/files#diff-016e53099417551f0595b3b00365c6083886f4e4208b555bb6901bc63a438f82R108-R109).

panglesd: (A type alias is a particular type declaration, where you are not introducing anything new, just giving a name to an existing type.

When you are defining a record, that is not a type alias: you are creating a new thing.
When you do `type t = <type expression>`, the type already existed and could be used by inlining the type expression.)

azzsal: Thanks for clarifying, I misunderstood a record type as being a type alias >:<

lib/diff.ml
```diff
@@ -107,34 +96,29 @@ let value_item ~typing_env ~name ~reference ~current =
...
-      | Module_item, reference, current ->
-          module_item ~typing_env:env ~name ~reference ~current
-      | Type_item, reference, current ->
-          type_item ~typing_env:env ~name ~reference ~current)
-    ref_items curr_items
-  |> Sig_item_map.bindings |> List.map snd
+  let diff_item : type a. (a, 'diff) Sig_item_map.diff_item =
+   fun item_type name reference current ->
+    match item_type with
+    | Sig_item_map.Value -> value_item ~typing_env:env ~name ~reference ~current
```
panglesd, line 107: Are those `Sig_item_map.` prefix path needed? The compiler should not need them, it knows the type of `item_type`. At least, the second and third one can be removed.

lib/diff.ml
```diff
@@ -107,34 +96,29 @@ let value_item ~typing_env ~name ~reference ~current =
...
-  | None, Some (Mod curr_md) ->
-      Some (Module { mname = name; mdiff = Added curr_md })
-  | Some (Mod ref_md), None ->
-      Some (Module { mname = name; mdiff = Removed ref_md })
-  | Some (Mod reference), Some (Mod current) ->
+  | None, Some curr_md -> Some (Module { mname = name; mdiff = Added curr_md })
+  | Some ref_md, None -> Some (Module { mname = name; mdiff = Removed ref_md })
+  | Some reference, Some current ->
       module_declaration ~typing_env ~name ~reference ~current
-  | _ -> assert false
```
panglesd, line 137: Very happy to see those wildcards and assert false go away!

lib/sig_item_map.ml
```diff
@@ -0,0 +1,77 @@
...
+type t = {
+  values_map : value_description String_map.t;
+  modules_map : module_declaration String_map.t;
+  types_map : (type_declaration * Ident.t) String_map.t;
+}
+
+type _ item_type =
+  | Value : value_description item_type
+  | Module : module_declaration item_type
+  | Type : (type_declaration * Ident.t) item_type
```
panglesd, line 12: There is a very little bit of type duplication here, that could be solved with a type alias, but we will do that if the types get more complex!

lib/sig_item_map.ml
```diff
@@ -0,0 +1,77 @@
...
+        values_map;
+        modules_map = String_map.add name item modules_map;
+        types_map;
+      }
+  | Type ->
+      {
+        values_map;
+        modules_map;
+        types_map = String_map.add name item types_map;
+      }
```
panglesd, line 41: Not destructing the input and using the `{ variable with field = ...}` syntax would make this function more concise.

lib/sig_item_map.ml
```diff
@@ -0,0 +1,77 @@
...
+    {
+      values_map = ref_values_map;
+      modules_map = ref_modules_map;
+      types_map = ref_types_map;
+    }
+    {
+      values_map = curr_values_map;
+      modules_map = curr_modules_map;
+      types_map = curr_types_map;
+    } : 'diff list =
```
panglesd, line 58: Again, I would find it more readable with this not destructed, especially since you are naming the fields.

panglesd (approved):
Great job. I left my nitpicks, but this can be merged before if needed.

(As I say, it might be better to wait for other PRs been merged)

NathanReb (approved):
This looks great! Thanks!

As @panglesd mentioned, might be worth holding this until some of the other PRs have been merged to prevent some rebase hell but other than that it's good to go!

CHANGES.md
```diff
@@ -8,6 +8,8 @@
 
 ### Changed
 
+- Abstracted Sig_item_map into is own module (#98, @azzsal)
```
NathanReb, line 11: This is internal so we don't need a changelog entry for it, I simply forgot to add the `no-changelog` label. You can safely remove it!

NathanReb:
Well done!

PR #100, opened 2024-10-21
Refactor value and type diffs

### LINKED ISSUE #99: Refactor `process_value_diff` and `process_type_diff`
(...one could factor the `process_value_diff` and `process_type_diff` and divide by two the number of lines they take by removing the duplicated logic.)

_Originally posted by @panglesd in https://github.com/ocaml-semver/ocaml-api-watch/pull/93#discussion_r1805926201_

NchamJosephMuam:
closes #99 

This PR refactors the functions that process value diffs and type diffs. 
I was able to write it after a short lesson from @panglesd about `GADTS`. Will appreciate any feedback.
@NathanReb @panglesd

### DIFF
--- a/lib/text_diff.ml
+++ b/lib/text_diff.ml
@@ -41,29 +41,24 @@ let md_to_lines name md =
   let module_str = "module " ^ name ^ ": " ^ Buffer.contents buf in
   CCString.lines module_str
 
-let process_value_diff (val_diff : Diff.value) =
-  match val_diff.vdiff with
-  | Added vd -> [ { orig = []; new_ = vd_to_lines val_diff.vname vd } ]
-  | Removed vd -> [ { orig = vd_to_lines val_diff.vname vd; new_ = [] } ]
+let process_diff :
+    type a.
+    (string -> a -> string list) ->
+    (a, a Diff.atomic_modification) Diff.t ->
+    _ ->
+    _ =
+ fun to_lines (diff : (a, a Diff.atomic_modification) Diff.t) name ->
+  match diff with
+  | Added item -> [ { orig = []; new_ = to_lines name item } ]
+  | Removed item -> [ { orig = to_lines name item; new_ = [] } ]
   | Modified { reference; current } ->
-      [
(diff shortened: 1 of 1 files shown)


NchamJosephMuam:
Please add the `no changelog` label

lib/text_diff.ml
```diff
@@ -41,29 +41,24 @@ let md_to_lines name md =
...
-let process_value_diff (val_diff : Diff.value) =
-  match val_diff.vdiff with
-  | Added vd -> [ { orig = []; new_ = vd_to_lines val_diff.vname vd } ]
-  | Removed vd -> [ { orig = vd_to_lines val_diff.vname vd; new_ = [] } ]
+let process_diff :
+    type a.
+    (string -> a -> string list) ->
+    (a, a Diff.atomic_modification) Diff.t ->
+    _ ->
+    _ =
```
panglesd, line 49: It is not required to give this type annotation to OCaml.

OCaml types structure items one by one. At the end of the typing of a structure, it may end up with some free variables: for instance, say, `'a -> 'a`. Then, it gives to the structure item the type : `'a. 'a -> 'a` which reads "For every type 'a, the structure item has the type 'a -> 'a` (so, the structure item has several types: `int -> int`, `string -> string`, ...).

Your annotation (which is slightly stronger than giving the type `'a . ...`) is therefore already inferred, and so unnecessary.

What OCaml's typer _does not_ do, is to give those "for every type `'a`, ..." _during_ the typing of a structure item, eg in a subexpression (in does not do that, because it is prove to be impossible to do).!
In your previous case, the function was recursive, so the structure item needed to be given the "forall" type _during_ the typing of it, and not only at the end. This is why it the type annotation was necessary.

Hope it explains sufficiently why you don't need a type annotation here!

NchamJosephMuam: Thank you for the explanation

panglesd (approved):
Looks good to me!

NathanReb (approved):

PR #102, opened 2024-10-28
Check Hidden values

### LINKED ISSUE #91: Ignore hidden values in modules
It can happen that modules have hidden values.

For instance, in the following example:

```
module type M = sig
  val x : int
end

module type N = sig
  include M
  val x : float
end
```
the `x` value coming from `include M` is shadowed by the "redeclaration" of `x`.

The visibility of a signature item is given by the [`visibility`](https://github.com/ocaml/ocaml/blob/trunk/typing/types.mli#L606) value of the [`signature_item`](https://github.com/ocaml/ocaml/blob/trunk/typing/types.mli#L626).

However, the API diffing currently ignores this visibility property.

This issue consists of two steps:

- Exhibit in a test a bug in the diffing that is due to this "visibility oversight"
- Fix it by taking the visibility into account when diffing two API.

NchamJosephMuam:
closes #91 
Hello @NathanReb @panglesd 
I added some code to explicitly check for visibility. I will now continue with writing the tests.

### DIFF
--- a/lib/diff.ml
+++ b/lib/diff.ml
@@ -19,18 +19,30 @@ type type_ = {
 
 type module_ = {
   mname : string;
-  mdiff : (module_declaration, module_modification) t;
+  mdiff : (module_declaration, signature_modification) t;
 }
 
-and module_modification = Unsupported | Supported of sig_item list
-and sig_item = Value of value | Module of module_ | Type of type_
+and modtype = {
+  mtname : string;
+  mtdiff : (modtype_declaration, signature_modification) t;
+}
 
-type item_type = Value_item | Module_item | Type_item [@@deriving ord]
+and signature_modification = Unsupported | Supported of sig_item list
+
+and sig_item =
+  | Value of value
+  | Module of module_
+  | Type of type_
+  | Modtype of modtype
+
+type item_type = Value_item | Module_item | Type_item | Modtype_item
+[@@deriving ord]
 
 type sig_items =
   | Val of value_description
   | Mod of module_declaration
   | Typ of type_declaration * Ident.t
+  | Modtype of modtype_declaration
 
 module Sig_item_map = Map.Make (struct
   type t = item_type * string [@@deriving ord]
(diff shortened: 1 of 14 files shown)


lib/diff.ml
```diff
@@ -38,18 +38,28 @@ and sig_item =
 let extract_items items =
   List.fold_left
     (fun tbl item ->
+      let visibility = Types.item_visibility item in
```
NathanReb, line 41: You don't need to call this function, the visibility is part of each sig item constructor arguments so you can directly pattern match over it.

lib/diff.ml
```diff
@@ -38,18 +38,28 @@ and sig_item =
...
       match item with
       | Sig_module (id, _, mod_decl, _, _) ->
-          Sig_item_map.add ~name:(Ident.name id) Sig_item_map.Module mod_decl
-            tbl
+          if visibility = Types.Hidden then tbl
```
NathanReb, line 44: Same here, you don't need to go through an if-then-else, you can directly match over the `visibility` argument (last one) and let the pattern matching wildcard case at the end handle all `Hidden` items.

lib/diff.ml
```diff
@@ -39,15 +39,15 @@ let extract_items items =
   List.fold_left
     (fun tbl item ->
       match item with
-      | Sig_module (id, _, mod_decl, _, _) ->
+      | Sig_module (id, _, mod_decl, _, Types.Exported) ->
```
NathanReb, line 42: The local open should not be necessary here! Same applies to the other cases.
```suggestion
      | Sig_module (id, _, mod_decl, _, Exported) ->
```

NathanReb (commented):
This looks good, thanks! Could you add a changelog entry in the `Fixed` section?

NchamJosephMuam:
> This looks good, thanks! Could you add a changelog entry in the `Fixed` section?
Yeah, thank you for the review. I'll add the changelog in the next commit

panglesd:
Since you merged `main` in this PR, there is a new case to handle in the pattern-match!

NchamJosephMuam:
> Since you merged `main` in this PR, there is a new case to handle in the pattern-match!

Thank you for catching that. I have updated it in the latest commit.

NathanReb (approved):

PR #103, opened 2024-11-19
Detect addition and removal of class type declarations

### LINKED ISSUE #83: Detect addition and removal of class type declarations

azzsal:
This aims to fix #83. Sorry about the delay.

There is a bit of code duplication with class declarations code. I will fix that, if you have an approach in mind, please let me know.

As with class declarations, a type is added when a class type declaration is added. Also a class type is added when a class declaration is added. This could be a bit confusing for a user of our tool/library. I think that should be fixed, but maybe by a different issue/PR.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@
 - Add detection of type declarations changes (#92, @azzsal)
 - Add detection of module_type declarations changes (#93, @NchamJosephMuam)
 - Add detection of classes addition and removal (#90, @marcndo)
-- Add detection of addition and removal of class type declarations (#92, @azzsal)
+- Add detection of addition and removal of class type declarations (#103, @azzsal)
 
 ### Changed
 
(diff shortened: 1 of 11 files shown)


azzsal:
@NathanReb 

I will open an issue later today to fix that. Thanks!

NathanReb (approved):

PR #105, opened 2024-12-05
Remove duplicate types and class types signature items

### LINKED ISSUE #104: Remove duplicate types and class types when detecting changes in class declarations and class type declarations.
As noted by #90 and #103, when adding a class declaration an object type and a class type are added by the compiler beside the class itself, and when adding a class type, an object type is added beside the class type. The same is true when removing or modifying a class or a class type. 

For example, if cls1 has been added to a module signature: 

```ocaml

class cls1 : object
  method m1 : int -> int
end
```

they will get this: 

```
 +type cls1 = < m1 : int -> int >
 +class cls1 : object method m1 : int -> int end
 +class type cls1 = object method m1 : int -> int end
```
whereby they expect just a class declaration is added:

```
 +class cls1 : object method m1 : int -> int end
```

I think a possible approach would be to check a field in the type_declaration, class_declaration or a class_type_declaration record before adding it to the Sig_item_map that we build at the start of the diffing process.

The tests should be updated accordingly as well.

azzsal:
This should resolve #104. I have dug down a bit in the ```Types``` module to find a record field that would distinguish a type from a class declaration, etc,  but I have not found any. I followed a simpler approach to not add a type if a class or a class type have been added, and not to add a class type if a class has been added. 

Sorry about the delay!

### DIFF
--- a/lib/sig_item_map.ml
+++ b/lib/sig_item_map.ml
@@ -42,18 +42,6 @@ let add (type a) ~name (item_type : a item_type) (item : a) maps : t =
         class_type_map = String_map.add name item maps.class_type_map;
       }
 
-let remove (type a) ~name (item_type : a item_type) maps : t =
-  match item_type with
-  | Value -> { maps with values_map = String_map.remove name maps.values_map }
-  | Module ->
-      { maps with modules_map = String_map.remove name maps.modules_map }
-  | Modtype ->
-      { maps with modtypes_map = String_map.remove name maps.modtypes_map }
-  | Type -> { maps with types_map = String_map.remove name maps.types_map }
-  | Class -> { maps with class_map = String_map.remove name maps.class_map }
-  | Classtype ->
-      { maps with class_type_map = String_map.remove name maps.class_type_map }
-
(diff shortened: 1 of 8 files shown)


lib/sig_item_map.ml
```diff
@@ -42,6 +42,27 @@ let add (type a) ~name (item_type : a item_type) (item : a) maps : t =
         class_type_map = String_map.add name item maps.class_type_map;
       }
 
+let remove (type a) ~name (item_type : a item_type) maps : t =
```
NathanReb, line 45: I believe this is unused, we can remove it for now!

NathanReb (commented):
This looks great! Thanks!

It does rely on the duplicates appearing after the class or class type but I don't think that's going to change so it's fine by me! We'll fix it if the need arises.

I have just one comment about the unused `remove` val but other than that it's good to go!

NathanReb:
Oh my bad, one last thing! Can you add a changelog entry under `Fixed`?

NathanReb (approved):
Awesome, thanks!

PR #106, opened 2025-01-01
Detect modification of class and class types

azzsal:
This is a continuation of #90 and #103. I added simple detection as other signature items.


NathanReb (commented):
This looks great, thanks!

We might need to do the `Includeclass` both ways, as in with reversed arguments as well.

Could you add a test or a pair of tests to make sure that something like:
```ocaml
class x : object
  method set : int -> int
end
```
and
```ocaml
class x : object
  method set : int -> int
  method size : int
end
```
are considered modified?

NathanReb (commented):
Besides the small typo, this looks good!

I'm surprised it does not consider one of the classes or class type in these new tests to be "more general", as specified by the compiler internal's documentation but I may have misunderstood it!

NathanReb:
I just realized I didn't hit the merge button on this, my bad. @azzsal do you want to take care of rebasing this or should I go ahead and do it?

NathanReb:
The CI's failing because one of the tests was not properly updated it seems, I'll fix this and merge!

PR #107, opened 2025-01-07
Add Support for unwrapped libraries

### LINKED ISSUE #71: Support unwrapped libraries
Once #70 has been fixed we should think about adding support for unwrapped libraries, i.e. libraries that don't have a single entry module.

azzsal:
This should eventually resolve #71. 
The following changes will be added incrementally:

- [x] Add a flag for unwrapped libraries in the api-diff CLI tool.
- [x] Add Library.load_unwrapped

### DIFF
--- a/bin/api_diff.ml
+++ b/bin/api_diff.ml
@@ -47,8 +47,10 @@ let run (`Main_module main_module) (`Unwrapped_library unwrapped)
     match curr_mode with
     | Wrapped main_module ->
         let main_module = String.capitalize_ascii main_module in
-        let+ reference_sig = Api_watch.Library.load ~main_module reference
-        and+ current_sig = Api_watch.Library.load ~main_module current in
+        let+ reference_map = Api_watch.Library.load ~main_module reference
+        and+ current_map = Api_watch.Library.load ~main_module current in
+        let reference_sig = Api_watch.String_map.find main_module reference_map in
+        let current_sig = Api_watch.String_map.find main_module current_map in
         let module_name = String.capitalize_ascii main_module in
         (reference_sig, current_sig, module_name)
     | Unwrapped -> failwith "TODO: Call Api_watch.Library.load_unwrapped"
--- a/bin/api_diff.ml
+++ b/bin/api_diff.ml
@@ -49,7 +49,9 @@ let run (`Main_module main_module) (`Unwrapped_library unwrapped)
         let main_module = String.capitalize_ascii main_module in
         let+ reference_map = Api_watch.Library.load ~main_module reference
         and+ current_map = Api_watch.Library.load ~main_module current in
-        let reference_sig = Api_watch.String_map.find main_module reference_map in
+        let reference_sig =
+          Api_watch.String_map.find main_module reference_map
+        in
         let current_sig = Api_watch.String_map.find main_module current_map in
         let module_name = String.capitalize_ascii main_module in
         (reference_sig, current_sig, module_name)
--- a/bin/api_diff.ml
+++ b/bin/api_diff.ml
@@ -10,29 +10,29 @@ let both_directories reference current =
       Error
         "Arguments must either both be directories or both single .cmi files."
 
+let print_warning main_module unwrapped =
+  match (main_module, unwrapped) with
+  | None, false -> ()
+  | Some _, false ->
+      Printf.eprintf
+        "%s: --main-module is ignored when diffing single .cmi files\n"
+        tool_name
+  | None, true ->
+      Printf.eprintf
+        "%s: --unwrapped is ignored when diffing single .cmi files\n" tool_name
+  | Some _, true ->
+      Printf.eprintf
+        "%s: --main-module and --unwrapped are ignored when diffing single \
+         .cmi files\n"
+        tool_name
+
 let mode ~reference ~current ~main_module ~unwrapped =
   match (both_directories reference current, main_module, unwrapped) with
   | Ok true, Some main_module, false -> Ok (Wrapped main_module)
   | Ok true, None, true -> Ok Unwrapped
-  | Ok false, main_module, unwrapped -> (
-      match (main_module, unwrapped) with
-      | None, false -> Ok Cmi
-      | Some _, false ->
-          Printf.eprintf
-            "%s: --main-module is ignored when diffing single .cmi files\n"
-            tool_name;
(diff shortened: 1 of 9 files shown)


bin/api_diff.ml
```diff
@@ -1,55 +1,42 @@
 let tool_name = "api-diff"
 
+type mode = Unwrapped | Wrapped of string | Cmi
+
+let rec mode ~reference ~current ~main_module ~unwrapped =
+  match (both_directories reference current, main_module, unwrapped) with
+  | Ok true, Some main_module, false -> Ok (Wrapped main_module)
+  | Ok true, None, true -> Ok Unwrapped
+  | Ok false, None, false -> Ok Cmi
+  | _ ->
```
panglesd, line 10: It could have been nice to split this case in several in order to give more specific message. But at the same time, a single message with all alternatives is fine!

bin/api_diff.ml
```diff
@@ -1,55 +1,42 @@
 let tool_name = "api-diff"
 
+type mode = Unwrapped | Wrapped of string | Cmi
+
+let rec mode ~reference ~current ~main_module ~unwrapped =
```
panglesd, line 5: This does not need to be recursive. `both_directories` simply needs to be defined first.

bin/api_diff.ml
```diff
@@ -1,55 +1,42 @@
...
+        "Inconsistent arguments. Either --main-module, --unwrapped-library or \
+         two single $(b,.cmi) files should be provided."
+
+and both_directories reference current =
+  match (Sys.is_directory reference, Sys.is_directory current) with
+  | true, true -> Ok true
+  | false, false -> Ok false
+  | _ ->
+      Error
+        "Arguments must either both be directories or both single .cmi files."
```
panglesd, line 21: The string in this error is defined but always ignored...

panglesd (commented):
Nice, the CLI file is much more readable like this!

bin/api_diff.ml
```diff
@@ -2,24 +2,43 @@ let tool_name = "api-diff"
...
+          Printf.eprintf
+            "%s: --unwrapped is ignored when diffing single .cmi files\n"
+            tool_name;
+          Ok Cmi
+      | Some _, true ->
+          Printf.eprintf
+            "%s: --main-module and --unwrapped are ignored when diffing single \
+             .cmi files\n"
+            tool_name;
+          Ok Cmi)
```
panglesd, line 35: You are returning `Ok Cmi` in each of this case. Factoring this out makes it more readable (we don't have to check each case to see what it's returning. Extract the warning printing in a function!

lib/library.ml
```diff
@@ -178,6 +178,19 @@ let rec expand_sig ~library_modules sig_ =
       | _ -> Ok item)
     sig_
 
+let load_unwrapped project_path =
+  let open CCResult.Infix in
+  let* library_modules = collect_modules project_path in
+  let module_map =
+    String_map.map
+      (fun sig_ ->
+        match expand_sig ~library_modules (Lazy.force sig_) with
```
NathanReb, line 187: Given that we'll load each individual module at the root here, I'm not quite sure we should expand them here. We could probably simply load the signature for each .cmi instead.

lib/library.mli
```diff
@@ -1,2 +1,5 @@
 val load_cmi : string -> (Types.signature * string, string) result
-val load : main_module:string -> string -> (Types.signature, string) result
+val load_unwrapped : string -> (Types.signature String_map.t, string) result
```
NathanReb, line 2: It's probably worth extracting a `Library.t` type here and to describe it a bit in the documentation .

In this case the map will contain all modules, but in the regular case, it will only contain one. That can be confusing to users so it should be documented that the map contains only the root modules of the library (i.e. all modules for unwrapped libs, only the main one for regular libs).

bin/api_diff.ml
```diff
@@ -42,31 +42,43 @@ let mode ~reference ~current ~main_module ~unwrapped =
 let run (`Main_module main_module) (`Unwrapped_library unwrapped)
     (`Ref_cmi reference) (`Current_cmi current) =
   let open CCResult.Infix in
-  let* reference_sig, current_sig, module_name =
+  let* reference_sig, current_sig =
```
NathanReb, line 45: These should probably renamed as they are not `signature` values anymore!

NathanReb (commented):
I didn't realize straight away that the PR wasn't ready yet, my bad!

Hope my comments can still help improve the PR in the meantime!

lib/library.ml
Siddhi-agg: Yeah, you are right! I made the necessary changes.

lib/library.mli
Siddhi-agg: Done

bin/api_diff.ml
Siddhi-agg: Done

bin/api_diff.ml
```diff
@@ -1,45 +1,82 @@
...
+  | Ok true, Some main_module, false -> Ok (Wrapped main_module)
+  | Ok true, None, true -> Ok Unwrapped
+  | Ok false, main_module, unwrapped ->
+      print_warning main_module unwrapped;
+      Ok Cmi
+  | (Error _ as e), _, _ -> e
+  | Ok true, _, _ ->
+      Error
+        "Either --main-module or --unwrapped must be provided when diffing \
+         entire libraries."
```
NathanReb, line 40: You can simplify the whole pattern-matching with a `let*` here:
```suggestion
  let* both_dirs = both_directories reference current in
  match (both_dirs, main_module, unwrapped) with
  | true, Some main_module, false -> Ok (Wrapped main_module)
  | true, None, true -> Ok Unwrapped
  | false, main_module, unwrapped ->
      print_warning main_module unwrapped;
      Ok Cmi
   | true, _, _ ->
      Error
        "Either --main-module or --unwrapped must be provided when diffing \
         entire libraries."
```

NathanReb (commented):
This looks good!

I wonder though if this is going to work in practice. Here we treat each module in isolation but it might be the case that they depend on each other and they needed to be diffed with a shared typing environment.

I'd suggest you open an issue to improve the handling of unwrapped libraries in a separate PR.

Aside from that and my simplification suggestion, this seems to be good to go!

Issue #108, opened 2025-01-07
Improve diff representation type declarations


NathanReb:
At the moment, the finer diff representation we have is per signature item. We can identify the different kinds of signature items and detect when one's added, removed or modified. In the latter case though, we don't go any deeper, meaning we simply represent the diff for said item as a pair of the old and new value.

The end goal of the tool is to instead have a very fine grained representation of the diff to immediatly be able to identify the precise parts of the API that changed.

Doing so for the whole .cmi representation is going to take time so we need to start small and type declarations seem like a good place to start.

## Type declarations diff

There are different components to type declarations that we'll need to be able to diff individually to describe how a type declaration was modified:
- Type privacy
- Type parameters
- Type manifest
- Concrete type definition for record or variant types

To keep things simple we could only implement this for either record or variant types at first and default to the current implementation if it's another kind of type declaration.

In a first iteration, we won't dive into fine grained diff of `core_type` values. Instead we can default to represent those as a `core_tpe * core_type` pair for the old and new value whenever they differ.

### What we don't need

Patching. We don't want to produce a patch as in something that we can use to construct the new API from the old one + the diff so that should not influence how we represent or optimize the diffing.

### What we need

We need to be able to display this diff into a text diff like format. Ideally we should be able to use it to also minimize this textual representation, i.e. to have something that could show a word based text diff when needed for instance.

We also want this representation to be easy to interpret programmatically, it will be exposed for consumption by other programs.

There will likely be a trade off here between a minimal representation or an independent one. I.e. are we able to print out the textual diff  from just the diff representation alone or do we need the original value + the diff. The minimal representation makes the diff easier to interpret programmatically but harder to print clearly on its own. This is still an open question and I think experimenting here will help us figure things out on this front.

NathanReb:
We can close this, great job @azzsal !

PR #109, opened 2025-01-10
Improve diff representation for record types

azzsal:
This PR implements a simple initial diff representation for record types. It aims to resolve a part of #108.
The following changes are made to the project:

- [x] Add a new diff representation for record types
- [x] using this representation when diffing record types
- [x] pretty printing the new representation as text (WIP)

### DIFF
--- a/lib/diff.ml
+++ b/lib/diff.ml
@@ -104,6 +104,11 @@ let module_type_fallback ~loc ~typing_env ~name ~reference ~current =
   | exception Includemod.Error _ ->
       Some (Module { mname = name; mdiff = Modified Unsupported })
 
+let extract_lbls lbls =
+  List.fold_left
+    (fun map lbl -> String_map.add (Ident.name lbl.ld_id) lbl map)
+    String_map.empty lbls
+
 let rec type_item ~typing_env ~name ~reference ~current =
   match (reference, current) with
   | None, None -> None
@@ -174,11 +179,6 @@ and modified_record_type ~ref_label_lst ~cur_label_lst =
     |> String_map.bindings |> List.map snd
   in
   (common_lbls, changed_lbls)
-
-and extract_lbls lbls =
-  List.fold_left
-    (fun map lbl -> String_map.add (Ident.name lbl.ld_id) lbl map)
-    String_map.empty lbls
 
 let value_item ~typing_env ~name ~reference ~current =
   match (reference, current) with
--- a/lib/diff.ml
+++ b/lib/diff.ml
@@ -12,8 +12,8 @@ type value = {
   vdiff : (value_description, value_description atomic_modification) t;
 }
 
-type ('common, 'changed) type_modification =
-  | Compound of 'common list * 'changed list
+type 'changed type_modification =
+  | Compound of 'changed list
   | Atomic of type_declaration atomic_modification
 
 and label_ = {
@@ -23,7 +23,7 @@ and label_ = {
 
 type type_ = {
   tname : string;
-  tdiff : (type_declaration, (label_declaration, label_) type_modification) t;
+  tdiff : (type_declaration, label_ type_modification) t;
 }
 
 type class_modification = Unsupported
@@ -130,15 +130,11 @@ let rec type_item ~typing_env ~name ~reference ~current =
       | _, _ -> (
           match (reference.type_kind, current.type_kind) with
           | Type_record (ref_label_lst, _), Type_record (cur_label_lst, _) ->
-              let common_lbls, changed_lbls =
+              let changed_lbls =
                 modified_record_type ~ref_label_lst ~cur_label_lst
               in
               Some
-                (Type
-                   {
-                     tname = name;
-                     tdiff = Modified (Compound (common_lbls, changed_lbls));
-                   })
+                (Type { tname = name; tdiff = Modified (Compound changed_lbls) })
           | _, _ ->
               Some
                 (Type
@@ -150,16 +146,6 @@ let rec type_item ~typing_env ~name ~reference ~current =
 and modified_record_type ~ref_label_lst ~cur_label_lst =
   let ref_lbls = extract_lbls ref_label_lst in
   let curr_lbls = extract_lbls cur_label_lst in
-  let common_lbls =
-    String_map.merge
-      (fun _ ref cur ->
-        match (ref, cur) with
-        | Some ref, Some cur ->
-            if ref.ld_type = cur.ld_type then Some ref else None
-        | _ -> None)
-      ref_lbls curr_lbls
-    |> String_map.bindings |> List.map snd
-  in
   let changed_lbls =
     String_map.merge
       (fun name ref cur ->
@@ -178,7 +164,7 @@ and modified_record_type ~ref_label_lst ~cur_label_lst =
       ref_lbls curr_lbls
     |> String_map.bindings |> List.map snd
   in
-  (common_lbls, changed_lbls)
+  changed_lbls
 
 let value_item ~typing_env ~name ~reference ~current =
   match (reference, current) with
--- a/lib/diff.ml
+++ b/lib/diff.ml
@@ -12,20 +12,16 @@ type value = {
   vdiff : (value_description, value_description atomic_modification) t;
 }
 
-type 'changed type_modification =
-  | Compound of 'changed list
+type type_modification =
+  | Compound of record_field list
   | Atomic of type_declaration atomic_modification
 
-and label_ = {
+and record_field = {
   lname : string;
   ldiff : (label_declaration, label_declaration atomic_modification) t;
 }
 
-type type_ = {
-  tname : string;
-  tdiff : (type_declaration, label_ type_modification) t;
-}
-
+type type_ = { tname : string; tdiff : (type_declaration, type_modification) t }
 type class_modification = Unsupported
 
 type class_ = {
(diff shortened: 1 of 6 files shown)


lib/diff.ml
panglesd, line 1: This is not introduced in this PR but I really don't think opening `Types` here is a good idea. It makes it very difficult to know where a type comes from: a definition above? Below? From `Types`?

Could you fix that? Either just use the full name, or if that is too long, use a shorter name (`module C = Type` (`C` for `Cmi`, but `T` could work, or `U` for upstream, just give the reason for the name as an attached comment!))

lib/diff.ml
```diff
@@ -4,19 +4,26 @@ type ('item, 'diff) t = Added of 'item | Removed of 'item | Modified of 'diff
...
 }
 
-type type_ = {
-  tname : string;
-  tdiff : (type_declaration, type_declaration atomic_modification) t;
+type type_modification =
+  | Record of record_modification list
+  | Any of type_declaration atomic_modification
+
+and record_modification = label_
```
panglesd, line 19: Why this additional indirection?

lib/diff.ml
```diff
@@ -113,8 +120,50 @@ let type_item ~typing_env ~name ~reference ~current =
...
+  let ref_lbls = extract_lbls ref_label_lst in
+  let curr_lbls = extract_lbls cur_label_lst in
+  String_map.merge
+    (fun name ref cur ->
+      match (ref, cur) with
+      | None, None -> None
+      | Some ref, None -> Some { lname = name; ldiff = Removed ref }
+      | None, Some cur -> Some { lname = name; ldiff = Added cur }
+      | Some ref, Some cur ->
+          if ref.ld_type = cur.ld_type then None
```
panglesd, line 153: I think we should be more careful here: The type may change without introducing a breaking change.

For instance:
```
type r = { a : int }
```
becoming
```
type a = int
type r = { a : a }
```

panglesd (commented):
Very nice!

lib/diff.ml
NathanReb: Yes definitely, this needs to be checked in a similar way to how we diff values!

Polymoprhic equality probably won't even work in the simple cases here anyway as the `core_type` representation is stateful and holds a lot of information beyond the type representation.

lib/diff.ml
```diff
@@ -113,8 +120,50 @@ let type_item ~typing_env ~name ~reference ~current =
...
+          else
+            Some
+              {
+                lname = name;
+                ldiff = Modified { reference = ref; current = cur };
+              })
+    ref_lbls curr_lbls
+  |> String_map.bindings |> List.map snd
+
+and extract_lbls lbls =
```
NathanReb, line 163: This can be extracted from the mutually recursive functions definition!

tests/api-diff/record_type_tests.t
```diff
@@ -0,0 +1,77 @@
...
+  
+Run the api-watcher on the two cmi files
+
+  $ api-diff ref.cmi modify_field_type.cmi 
+  diff module Modify_field_type:
+  *type student = {
+    first_name: string;
+    last_name: string;
+    -id: int option
+    +id: int
```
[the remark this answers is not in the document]
NathanReb: Yes that's an idea! I wonder if we could something a bit more inlined, maybe by having some mechanism to highlight the part of a line that changed in the colored output.

i.e. by having:
```
type student = {
    first_name: string;
    last_name: string;
-   id: int option
+   id: int
  }
```

Where option has a different background color, or maybe just:
```
type student = {
    first_name: string;
    last_name: string;
-   id: int option
  }
```
With only option shown in red and the rest of the line in the default color.

Would have to look up how other tools display word based diff to take inspiration from them!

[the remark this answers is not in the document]
NathanReb: I kind of like the github way to show this:
Screenshot from 2025-01-13 11-28-21

lib/diff.ml
azzsal: If I am not mistaken, there should be a way to lookup a `type_expr` in the typing_env, and check if the alias resolves to the same type, right ?

lib/text_diff.ml
```diff
@@ -1,21 +1,19 @@
 open Types
 
-type change = { orig : string list; new_ : string list }
-type conflict2 = Atomic of change | Record of string * change list
-type t = conflict2 list String_map.t
-type printer = { same : string Fmt.t; diff : conflict2 Fmt.t }
+type conflict2 = { orig : string list; new_ : string list }
+type common = string list
+type context = Change of conflict2 | Same of common
```
NathanReb, line 3: `context` is a bit confusing here as it can both be the context or the actual change. The terminology in the original `diffutils` for this was `hunk`, we could eventually re-use this name.

lib/diff.ml
```diff
@@ -12,18 +12,20 @@ type value = {
   vdiff : (value_description, value_description atomic_modification) t;
 }
 
-type type_modification =
-  | Record of record_modification list
-  | Any of type_declaration atomic_modification
-
-and record_modification = label_
+type ('common, 'changed) type_modification =
+  | Compound of 'common list * 'changed list
```
NathanReb, line 16: It feels like we're adding the unchanged labels into our diff representation because we need them to print context but not because we need them to represent the diff.

Not all consumers of the diff will need this so we need to really consider why we're adding it here.

Is there a simple way we could produce the text diff from both the original signature and the diff for instance?
My intuition is that with just the signature it would be a bit redundant but we can maybe pass an abstraction on top of signature, something like the item map would probably be easier to use there. That's something we could consider.

Another thing to take into account here: what happens if we have a single diff in a very large record. We'll embed all unchanged fields even though in `Text_diff` we'll probably want to only print a few of them for context, something like:
```
 type t =
   { first_field : int
 ...
   ; n_minus_one_field : int
-  ; n_field : int
+  ; n_field : string
   ; n_plus_one_field : bool
```

What would we do then? Should we embed all unchanged fields? Only the one we use for the text diff? I don't think we'd have a very clear guideline on what to embed and why.

I have the impression that the `Text_diff` implementation is leaking into `Diff`.

My gut feeling is that the diff being minimal is probably the best solution, we can provide the context to `Text_diff` ourselves.

I imagine we could even have two different versions of the text diff representation, one that prints some context and one that doesn't.

What I'd suggest here is the following: let's stick to the minimal diff representation for now. We make `Text_diff` simply print the changed parts without context in this PR and leave the context printing for a separate PR.

What do you think?

azzsal: I kinda felt the same way in regard to the implementation of the two modules implementation were leaking into each other. I was looking into the `Diff` module and `Text_diff` modules at the same, and trying to match one to the other. I think that violates the principle of encapsulation in some sense. 

Having two different text diff representation based on some criteria seem a good option too. Maybe that should be the goal for the next PR. I try and look at different word-based diffing tools to take some inspirations from them.

I think I will go with a minimal representation without the context as you suggested as something to start with.

lib/diff.ml
```diff
@@ -4,17 +4,26 @@ type ('item, 'diff) t = Added of 'item | Removed of 'item | Modified of 'diff
...
-     Use this type when there is no better representation available. *)
+    [reference] is the value before and [current] is the value after the change
+    occured. Use this type when there is no better representation available. *)
 
 type value = {
   vname : string;
   vdiff : (value_description, value_description atomic_modification) t;
 }
 
+type 'changed type_modification =
```
NathanReb, line 15: Is there a specific reason why this is polymorphic?

My guess is that it's to leave room for the variant thing but I think we could still remove the parameter as we'll likely have to use some sort of:
```
| Variant of constructor list
| Record_diff of label_ list
...
```

Since the `type_declaration` type itself is not polymorphic, I don't think we need to be here but I might be missing something.

lib/diff.ml
```diff
@@ -4,17 +4,26 @@ type ('item, 'diff) t = Added of 'item | Removed of 'item | Modified of 'diff
...
+  | Compound of 'changed list
+  | Atomic of type_declaration atomic_modification
+
+and label_ = {
```
NathanReb, line 19: I don't think label is a keyword so you should be able to go for `label` if you'd like.

That being said, label is used for other things as well in the compiler libs so we could disambiguate this by making it `record_field`. What do you think?

lib/diff.ml
```diff
@@ -113,8 +127,45 @@ let type_item ~typing_env ~name ~reference ~current =
...
+          | _, _ ->
+              Some
+                (Type
+                   {
+                     tname = name;
+                     tdiff = Modified (Atomic { reference; current });
+                   })))
+
+and modified_record_type ~typing_env ~ref_label_lst ~cur_label_lst =
+  let _ = typing_env in
```
NathanReb, line 147: You can remove this as now `typing_env` is used!
```suggestion
```

NathanReb (commented):
This looks great, thanks for your work on this!

Aside from my minor comments this is good to go!

lib/diff.ml
azzsal: `record_field` seems a better name.

lib/diff.ml
azzsal: Yes, I was considering that for the variant case. I think doing it as the compiler does it might be a possibility too. 
`type_decl_kind = (label_declaration, constructor_declaration) type_kind`

I will think about that once I rebase the variant PR.

NathanReb (approved):
Looks good, thanks!

PR #111, opened 2025-01-14
Improve diff representation for variant types

azzsal:
This will also be a part of the exploration to have a better diff representation for type declarations.  
The new diff representation will capture concrete variant type declarations. 
It is part of the work on #108.

The following changes will be made in this PR:

- [x] Adding a new diff representation for variant types
- [x] Using this representation when diffing variant types
- [x] Pretty printing the new representation as text diff


azzsal:
Sorry, I branched off the wrong branch again, I will try and fix that tomorrow.

azzsal:
There's some duplication with record types PR here, but I think that's something that we could take care of once we rebase/merge.

lib/diff.ml
```diff
@@ -130,32 +150,133 @@ let rec type_item ~typing_env ~name ~reference ~current =
...
+              let changed_constrs =
+                modified_variant_type ~typing_env ~ref_constructor_lst
+                  ~cur_constructor_lst
+              in
+              Some
+                (Type
+                   {
+                     tname = name;
+                     tdiff = Modified (Variant_diff changed_constrs);
+                   })
```
NathanReb, line 166: This is not entirely guaranteed to contain a diff. For instance, one could change the variance of one the type parameters without changing the concrete definition. There would be a diff for the type declaration but the individual contructors's definition would remain identical.

I think this also holds for the record case above.

Could you add a test that exhibits that, make sure that the current implementation fails said test and fix it?

tests/api-diff/modified_variant_type_tests.t
```diff
@@ -0,0 +1,155 @@
...
+
+  $ ocamlc modify_field.mli 
+
+Run the api-watcher on the two cmi files
+
+  $ api-diff shapes.cmi modify_field.cmi
+  diff module Modify_field:
+   type shape =
+    ...
+    | Circle of { ...; -raduis : int; +raduis : float; }
```
panglesd, line 153: ?
```suggestion
    | Circle of { ...;
-      raduis : int;
+      raduis : float;
  }
```
or for another time?
```suggestion
    | Circle of { ...;
       raduis :
-       int;
+       float;
  }
```

lib/text_diff.ml
```diff
@@ -1,174 +1,245 @@
 type conflict2 = { orig : string list; new_ : string list }
-type hunk = Change of conflict2 | Same of string
+type hunk = Change of conflict2 * bool | Same of string * bool
```
panglesd, line 2: Arguments whose role is not explicit from the type should always be named. You can use a record here to name the fields!

lib/diff.ml
```diff
@@ -123,39 +143,169 @@ let rec type_item ~typing_env ~name ~reference ~current =
...
-          | Type_record (ref_label_lst, _), Type_record (cur_label_lst, _) ->
-              let changed_lbls =
-                modified_record_type ~typing_env ~ref_label_lst ~cur_label_lst
-              in
-              Some
-                (Type { tname = name; tdiff = Modified (Compound changed_lbls) })
-          | _, _ ->
+          | Type_record (ref_label_lst, _), Type_record (cur_label_lst, _) -> (
+              match (ref_mismatch, cur_mismatch) with
+              | ( Some (Includecore.Record_mismatch _),
```
NathanReb, line 150: I'm not sure that's the right approach here, what if the types differ on multiple levels and Includecore returns one of the other reasons? We'd still want to show that diff.

We have to trust that our function does the right thing. Compute the diff between the two records, if it's `[]` then return `None`, otherwise returns the list of diffs!

lib/diff.ml
```diff
@@ -123,39 +143,169 @@ let rec type_item ~typing_env ~name ~reference ~current =
...
                      tdiff = Modified (Atomic { reference; current });
                    })))
 
-and modified_record_type ~typing_env ~ref_label_lst ~cur_label_lst =
+and modified_variant_type ~typing_env ~ref_constructor_lst ~cur_constructor_lst
+    =
+  let same_args cstr1_args cstr2_args =
+    match (cstr1_args, cstr2_args) with
+    | Cstr_tuple type_lst1, Cstr_tuple type_lst2 -> (
+        try
```
NathanReb, line 192: Why are you using a `try` block here? Is it because `Ctype.does_match` can throw? If that's the case, I think we should extract a function of our own to diff two `type_expr` that wraps `does_match` and deals with the exception correctly. That way we can reuse it across.

Also, it's worth noting that we try to avoid `with _ ->` blocks as we could catch an unexpected exception coming from further down the line and that should not be interpreted as "the types differ". If we expect `does_match` to throw an exception that tells us the types differ, we should catch only that exception or that specific set of exceptions.

lib/diff.ml
```diff
@@ -123,39 +143,169 @@ let rec type_item ~typing_env ~name ~reference ~current =
...
+          if
+            List.for_all2
```
NathanReb, line 194: We're going through the entire list twice here: once to tell if there is a diff at all, and a second time to compute the actual diff, calling `does_match` on all elements of the list each time.

I think we can simplify all this by simply computing the diff as a `diff option list` first and simply detecting the case where it's all `None`. That way we call `does_match` only once for each pair.

lib/diff.ml
NathanReb: This applies for the record case below as well.

lib/diff.ml
NathanReb: This applies for the record case below as well.

lib/diff.ml
NathanReb: This applies for the record case below as well.

lib/text_diff.ml
```diff
@@ -40,6 +43,20 @@ let lbl_to_lines ld =
   Format.pp_print_flush formatter ();
   CCString.lines (Buffer.contents buf)
 
+let te_to_lines te =
```
NathanReb, line 46: what does `te` stands for here? Is it tuple?

tests/api-diff/modified_variant_type_tests.t
```diff
@@ -0,0 +1,261 @@
...
+Run the api-watcher on the two cmi files
+
+  $ api-diff ref.cmi modify_constructor_type.cmi
+  diff module Modify_constructor_type:
+   type rank =
+    ...
+  - | Number of (
+  -    int
+  +    float
+    )
```
NathanReb, line 72: I think this would read better as:
```
- | Number of int
+ | Number of float
```

tests/api-diff/modified_variant_type_tests.t
```diff
@@ -0,0 +1,261 @@
...
+
+  $ ocamlc remove_field.mli 
+
+Run the api-watcher on the two cmi files
+
+  $ api-diff shapes.cmi remove_field.cmi
+  diff module Remove_field:
+   type shape =
+    ...
+    | Circle of { ...; -raduis : int; }
```
NathanReb, line 138: It seems this one has been fixed yet!

tests/api-diff/modified_variant_type_tests.t
NathanReb: Yes this should be simplified, I think the one line per field is the right approach for now!

NathanReb (commented):
This is starting to look good. I think there are quite a bit that can be simplified here but it's a good step in the right direction. I'm looking forward to seeing the simplified version!

tests/api-diff/modified_variant_type_tests.t
panglesd, line 72: The `-` before the constructor name should be removed!
Also, I find the parentheses a bit weird
(And they might interfere with the OCaml syntax. For instance,
```
type t1 = A of int * int
type t2 = A of (int * int)
```
are valid OCaml syntax which mean different things: one has the tuple "unboxed" while the other has it boxed: see `utop -dparsetree` !)

lib/diff.ml
azzsal: I am using `try` because `List.for_all2` raises Invalid_argument exception if the the two `type_expr lists` weren't of the same length, in which case they aren't identical, and I call `modified_tuple_type` function to determine their diff. 
`Ctype.does_match` doesn't raise, `Ctype.matches` does raise however, `Ctype.does_match` simply calls `Ctype.matches` and return false in the case that it raises a `Matches_failure`.

> Also, it's worth noting that we try to avoid with _ -> blocks as we could catch an unexpected exception coming from further down the line and that should not be interpreted as "the types differ".

Yes I agree, sorry my bad. A try should be wrapped around the closest expression that we expect to raise the exception, but I think I will remove `List.for_all2` all together as you pointed down below.

lib/text_diff.ml
azzsal: It stands for type_expr. Sorry but I followed the same naming convention as in the other to_lines functions :)

tests/api-diff/modified_variant_type_tests.t
```diff
@@ -0,0 +1,262 @@
...
+
+  $ api-diff shapes2.cmi add_component.cmi
+  diff module Add_component:
+   type shape =
+    ...
+    | Circle of
+       point *
+       int *
+  +    int
+    | Rectangle of
```
NathanReb, line 206: I still believe this should be displayed on a single line!

.ocamlformat
```diff
@@ -1 +1 @@
-version=0.26.2
```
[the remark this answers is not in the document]
NathanReb: Can you remove this change please?

NathanReb (commented):
There are some improvement but this can still be simplified a lot.

Let's display variant with tuple arguments change on a single line for now.

lib/diff.ml
```diff
@@ -125,37 +145,153 @@ let rec type_item ~typing_env ~name ~reference ~current =
...
+        | None, None -> None
+        | Some ref, None -> Some { csname = name; csdiff = Removed ref }
+        | None, Some cur -> Some { csname = name; csdiff = Added cur }
+        | Some ref, Some cur -> diff_cstrs name ref cur)
+      ref_cstrs curr_cstrs
+    |> String_map.bindings |> List.map snd
+  in
+  modified_cstrs
+
+and modified_tuple_type ~typing_env (ref_tuple : type_expr list)
```
NathanReb, line 230: This function can be simplified a lot. I think we'll run into similar cases in the future so I suggest extracting a generic:
```ocaml
val diff_list :
  dif_one: (ref: 'a option -> current: 'a option -> ('a, 'diff) Either.t) ->
  ref: 'a list ->
  current: 'a list ->
  ('a, 'diff) Either.t list
```
Where `diff_one` would receive `None` when there is no corresponding element in the relevant list because their length differ.

This can be implemented in a single traversal of both list. Let me know if you need help writing this.

Once you have that generic function, writing `modified_tuple_type` should be trivial.

lib/diff.ml
```diff
@@ -125,37 +145,153 @@ let rec type_item ~typing_env ~name ~reference ~current =
...
+and modified_variant_type ~typing_env ~ref_constructor_lst ~cur_constructor_lst
+    =
+  let diff_cstrs name cstr1 cstr2 =
+    match (cstr1.cd_args, cstr2.cd_args) with
+    | Cstr_tuple type_lst1, Cstr_tuple type_lst2
+      when not
+             ((List.length type_lst1 = 1 && List.length type_lst2 = 1)
+             && not
+                  (Ctype.does_match typing_env (List.hd type_lst1)
+                     (List.hd type_lst2))) ->
```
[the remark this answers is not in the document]
azzsal: Well, you [asked](https://github.com/ocaml-semver/ocaml-api-watch/pull/111#discussion_r1923437945) for it up above :) 
When a constructor argument is a tuple and it has a cardinality of one and has been modified it should be an atomic diff.

tests/api-diff/modified_variant_type_tests.t
azzsal: Shouldn't the inling of text diffs be done in a separate PR ?

lib/diff.ml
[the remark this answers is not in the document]
NathanReb: Ok my bad. What I was suggesting is that we display constructor with tuple args diff on a single line, no matter the number of arguments. We still want the nice diff representation internally, my request was simply a textual diff related matter.

tests/api-diff/modified_variant_type_tests.t
NathanReb: I'm not suggesting any highlighting here, simply that the line based diff is showed as:
```
-  | Circle of point * int
+  | Circle of point * int * int
```
as that is simply more readable.

lib/text_diff.ml
```diff
@@ -103,39 +155,80 @@ let process_lbl_diff
...
+
+and process_modified_tuple_type_diff name diff =
+  let starify typ_opts =
+    let typs =
+      List.fold_right
+        (fun t_opt acc ->
+          match t_opt with
+          | None -> acc
+          | Some t -> String.concat "" (typ_expr_to_lines t) :: acc)
+        typ_opts []
```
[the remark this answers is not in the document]
NathanReb: Also note that you're using type_expr_to_lines which split out the string into a list of strings to then recombine it. It might be a better approach to have a function that directly produces a one line string for a given type_expr. Have you tried using Printtyp.type_expr directly?

If that does not work or to simply err on the safe side, you can extract a function that turns a multi-line string into a single line one, e.g. by mapping `'\n'` to `' '`. You can then use this along with `Printtyp.type_expr` to directly get a one line string.

tests/api-diff/record_type_tests.t
```diff
@@ -92,3 +92,21 @@ Run api-watcher on the two cmi files
...
+
+We generate the .cmi file
+
+  $ ocamlc add_type_parameter.mli << EOF
+
+Run api-watcher on the two cmi files
+
+  $ api-diff ref.cmi add_type_parameter.cmi
+  diff module Add_type_parameter:
+  +<unsupported change>
```
NathanReb, line 110: This should not be reported as an unsupported change here, there's something fishy going on.

I'd expect this to be represented as an atomic type_decl diff and therefore dispayed as such.

The problem might come from somewhere else. Could you try to see if adding this test on top of main shows the same result?

It's probably best to take that test out of this PR anyway as it was either always there or was introduced in the record diffing PR.

NathanReb (commented):
This is looking much better, thank!

There are still a few simplifications left but I think after those it's good to go!

tests/api-diff/record_type_tests.t
azzsal: It's also an unsupported change on main, but also there like an output of an changed record type declaration.

NathanReb: Can you extract this test out of this PR and make a new PR on top of main with just that test? we should fix this separately?

lib/text_diff.ml
```diff
@@ -43,12 +43,12 @@ let lbl_to_lines ld =
...
   CCString.lines (Buffer.contents buf)
 
-let typ_expr_to_lines typ_exp =
+let typ_expr_to_line typ_exp =
   let buf = Buffer.create 256 in
   let formatter = Format.formatter_of_buffer buf in
   Printtyp.type_expr formatter typ_exp;
   Format.pp_print_flush formatter ();
-  CCString.lines (Buffer.contents buf)
+  String.map (fun c -> if c = '\n' then ' ' else c) (Buffer.contents buf)
```
NathanReb, line 51:
```suggestion
  String.map (function '\n' -> ' ' | c -> c) (Buffer.contents buf)
```

tests/api-diff/record_type_tests.t
NathanReb: my bad, you already did!

NathanReb (approved):
Thanks for your hard work on this, this looks great!

PR #113, opened 2025-01-27
Handling type variables in type declaration diffs

### LINKED ISSUE #112: Handle type parameters when diffing record and varaint types
Our current implementation only takes care of the concrete type declaration for record and variant types when diffing them respectively.

The end goal is to have fine grain diffs of type declarations.
Having a diff between type parameters  will get us one step closer.

The issue will take care of the following:

**Adding a type parameter:**

```
type t = A of int 

type 'a t = A of 'a

```
Suggested text diff representation:

```
-type t =
+type t1 t =
- | A of int
+ | A of 'a
```

**Removing a type parameter:**

```
type 'a t = A of 'a 

type t = A of int

```
Suggested text diff representation:

```
-type t1 t =
+type t =
- | A of t1
+ | A of int
```

Renaming type variables shouldn't be considered a diff, having something like this shouldn't report a diff:

```
type 'a t = A of 'a 

type 'b t = A of 'b
```

azzsal:
This should resolve #112.

### DIFF
--- a/lib/diff.ml
+++ b/lib/diff.ml
@@ -163,13 +163,11 @@ let rec type_item ~typing_env ~name ~reference ~current =
       type_declarations ~typing_env ~name ~reference ~current
 
 and type_declarations ~typing_env ~name ~reference ~current =
-  let reference, current =
-    if
-      Normalize.is_type_params ~reference:reference.Types.type_params
-        ~current:current.Types.type_params
-    then (reference, current)
-    else Normalize.type_declarations ~reference ~current
-  in
+  if
+    Normalize.is_type_params ~reference:reference.Types.type_params
+      ~current:current.Types.type_params
+  then ()
+  else Normalize.type_declarations ~reference ~current;
   let ref_params = reference.type_params in
   let cur_params = current.type_params in
   let type_kind =
--- a/lib/diff.ml
+++ b/lib/diff.ml
@@ -132,25 +132,27 @@ let module_type_fallback ~loc ~typing_env ~name ~reference ~current =
 
 let type_expr typing_env reference current =
   let rec same_type_var_names reference current =
-    match get_desc reference, get_desc current with
+    match (get_desc reference, get_desc current) with
     | Tvar None, Tvar None -> true
-    | Tvar (Some ref_name), Tvar (Some cur_name) -> 
-      String.equal ref_name cur_name
-    | Tarrow (_, ref_te1, ref_te2, _),
-      Tarrow (_, cur_te1, cur_te2, _) ->
-      (same_type_var_names ref_te1 cur_te1) && 
-      (same_type_var_names ref_te2 cur_te2)
-    | Ttuple (ref_te_lst), Ttuple(cur_te_lst) -> 
-      if List.length ref_te_lst != List.length cur_te_lst then false
-      else List.for_all2 (fun ref_te cur_te -> same_type_var_names ref_te cur_te)
-          ref_te_lst cur_te_lst
-   | _ -> true
+    | Tvar (Some ref_name), Tvar (Some cur_name) ->
+        String.equal ref_name cur_name
+    | Tarrow (_, ref_te1, ref_te2, _), Tarrow (_, cur_te1, cur_te2, _) ->
+        same_type_var_names ref_te1 cur_te1
+        && same_type_var_names ref_te2 cur_te2
+    | Ttuple ref_te_lst, Ttuple cur_te_lst ->
+        if List.length ref_te_lst != List.length cur_te_lst then false
+        else
+          List.for_all2
+            (fun ref_te cur_te -> same_type_var_names ref_te cur_te)
+            ref_te_lst cur_te_lst
+    | _ -> true
   in
-  let equal = Ctype.does_match typing_env reference current
-      && same_type_var_names reference current
+  let equal =
+    Ctype.does_match typing_env reference current
+    && same_type_var_names reference current
   in
   if equal then None else Some (Modified { reference; current })
-        
+
 let extract_lbls lbls =
   List.fold_left
     (fun map lbl -> String_map.add (Ident.name lbl.ld_id) lbl map)
--- a/lib/diff.ml
+++ b/lib/diff.ml
@@ -131,14 +131,26 @@ let module_type_fallback ~loc ~typing_env ~name ~reference ~current =
       Some (Module { mname = name; mdiff = Modified Unsupported })
 
 let type_expr typing_env reference current =
-  let equal =
-    match (get_desc reference, get_desc current) with
-    | Tvar (Some ref_name), Tvar (Some cur_name) ->
-        Ctype.does_match typing_env reference current && ref_name = cur_name
-    | _, _ -> Ctype.does_match typing_env reference current
+  let rec same_type_var_names reference current =
+    match get_desc reference, get_desc current with
+    | Tvar None, Tvar None -> true
+    | Tvar (Some ref_name), Tvar (Some cur_name) -> 
+      String.equal ref_name cur_name
+    | Tarrow (_, ref_te1, ref_te2, _),
+      Tarrow (_, cur_te1, cur_te2, _) ->
+      (same_type_var_names ref_te1 cur_te1) && 
+      (same_type_var_names ref_te2 cur_te2)
+    | Ttuple (ref_te_lst), Ttuple(cur_te_lst) -> 
+      if List.length ref_te_lst != List.length cur_te_lst then false
+      else List.for_all2 (fun ref_te cur_te -> same_type_var_names ref_te cur_te)
+          ref_te_lst cur_te_lst
+   | _ -> true
+  in
+  let equal = Ctype.does_match typing_env reference current
+      && same_type_var_names reference current
   in
   if equal then None else Some (Modified { reference; current })
-
+        
 let extract_lbls lbls =
   List.fold_left
     (fun map lbl -> String_map.add (Ident.name lbl.ld_id) lbl map)
@@ -157,13 +169,16 @@ let rec type_item ~typing_env ~name ~reference ~current =
   | None, Some (current, _) ->
       Some (Type { tname = name; tdiff = Added current })
   | Some (reference, _), Some (current, _) ->
-      let normalized_ref_type_decl, normalized_cur_type_decl =
-        Util.normalize_type_decls reference current
-      in
-      type_decls ~typing_env ~name ~reference:normalized_ref_type_decl
-        ~current:normalized_cur_type_decl
+      type_decls ~typing_env ~name ~reference ~current
 
 and type_decls ~typing_env ~name ~reference ~current =
+  let reference, current =
+    if
+      Util.normalized_type_params reference.Types.type_params
+        current.Types.type_params
+    then (reference, current)
+    else Util.normalize_type_decls reference current
+  in
   let type_kind =
     type_kind ~typing_env ~reference:reference.type_kind
       ~current:current.type_kind
--- a/lib/diff.ml
+++ b/lib/diff.ml
@@ -130,28 +130,17 @@ let module_type_fallback ~loc ~typing_env ~name ~reference ~current =
   | exception Includemod.Error _ ->
       Some (Module { mname = name; mdiff = Modified Unsupported })
 
-let type_expr typing_env reference current =
-  let rec same_type_var_names reference current =
-    match (get_desc reference, get_desc current) with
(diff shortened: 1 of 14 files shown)


azzsal:
I have added normalization of type parameters in type declarations. I still haven't normalized the type variables in the type_manifest and type_kind. I am not sure if I am moving in the right direction.
Could you review the the `normalize_type_decls` in `lib/diff.ml` ? @NathanReb @panglesd

lib/diff.ml
```diff
@@ -149,7 +157,207 @@ let rec type_item ~typing_env ~name ~reference ~current =
...
+        ( normalized_ref_type_param :: normalized_ref_type_params',
+          normalized_cur_type_param :: normalized_cur_type_params',
+          String_map.add ref_type_param_name unique_name ref_type_var_map',
+          String_map.add cur_type_param_name unique_name cur_type_var_map' )
+  in
+  let ( normalized_ref_type_params,
+        normalized_cur_type_params,
+        ref_type_var_name_map,
+        cur_type_var_name_map ) =
+    normalize_type_params
```
panglesd, line 237: This function should be split in two:
- The part that builds the type two maps
- The part that normalize a type params

lib/diff.ml
```diff
@@ -149,7 +157,207 @@ let rec type_item ~typing_env ~name ~reference ~current =
...
+  in
+  let normalized_cur_type_manifest =
+    normalize_type_manifest cur_type_decl.type_manifest cur_type_var_name_map
+  in
+  let normalized_ref_type_kind =
+    normalize_type_kind ref_type_decl.type_kind ref_type_var_name_map
+  in
+  let normalized_cur_type_kind =
+    normalize_type_kind cur_type_decl.type_kind cur_type_var_name_map
+  in
```
panglesd, line 252: You should avoid repeating code! In this case, define a `normalize_type_decl` which takes a type decl and a map, and call it twice for `ref` and `current`!

lib/diff.ml
```diff
@@ -149,7 +157,207 @@ let rec type_item ~typing_env ~name ~reference ~current =
...
+      type_kind = normalized_cur_type_kind;
+    } )
+
+and normalize_type_expr tvmap type_expr =
+  match get_desc type_expr with
+  | Tvar None -> type_expr
+  | Tvar (Some name) ->
+      let nname = String_map.find name tvmap in
+      create_expr (Tvar (Some nname)) ~level:(get_level type_expr)
+        ~scope:(get_scope type_expr) ~id:(get_id type_expr)
```
panglesd, line 272: Define a `create_expr` function with less arguments, which is just a wrapper to `create_expr`!

lib/text_diff.ml
```diff
@@ -206,20 +234,34 @@ and pair_of_manifest_diff manifest_diff =
...
+    private_flag_str manifest_str
 
 and string_of_private_flag private_flag =
   match private_flag with Public -> "" | Private -> " private"
 
 and string_of_manifest manifest =
   match manifest with None -> "" | Some te -> " " ^ typ_expr_to_line te
 
+and string_of_type_params (type_params : Types.type_expr list) =
+  let wrap = if List.length type_params > 1 then true else false in
```
panglesd, line 254: Really nitpicking but `if ... then true else false` can always be simpilified!
```suggestion
  let wrap = List.length type_params > 1 in
```
Nitpicking even more, I would do it like that to avoid computing the length:
```suggestion
  let wrap = match type_params with _ :: _ :: _ -> true | _ -> false in
```

lib/text_diff.ml
```diff
@@ -206,20 +234,34 @@ and pair_of_manifest_diff manifest_diff =
...
+  (if wrap then "(" else "")
+  ^ String.concat ", " (List.map get_type_param_name type_params)
+  ^ (if wrap then ") " else "")
+  ^ if List.length type_params = 1 then " " else ""
```
panglesd, line 258: again you can use a match to avoid computing the length!
```suggestion
  ^ (match type_params with [ _ ] -> " " | _ ->) ""
```

lib/diff.ml
```diff
@@ -128,8 +131,13 @@ let module_type_fallback ~loc ~typing_env ~name ~reference ~current =
...
 
 let type_expr typing_env reference current =
-  if Ctype.does_match typing_env reference current then None
-  else Some (Modified { reference; current })
+  let equal =
+    match (get_desc reference, get_desc current) with
+    | Tvar (Some ref_name), Tvar (Some cur_name) ->
+        Ctype.does_match typing_env reference current && ref_name = cur_name
+    | _, _ -> Ctype.does_match typing_env reference current
+  in
```
panglesd, line 139: Why do you make a special case here?
For instance, if `reference` is `'a` and `current` is `'b` we get the special case, but if `reference` is `'a -> 'a` and `current` is `'b -> 'b` we do not!

This might be a problem...

panglesd (commented):
Very nice!

The most meaningful comment is about the special case when diffing type expressions...

The rest of my comments are about refactoring to make it cleaner/shorter/easier to read/less error prone. Quite easy to fix but I think together they will improve the readability!

One last comment is that I find it a shame that we lose the handwritten type params _even_ when there is no changes, which is most of the time... After you do the refactoring, could you normalize only if there is a change in type params?

lib/diff.ml
azzsal: This is the "hack" I depended on instead of checking the binding of a type variable.. which I wasn't sure how to do that, but @NathanReb  mentioned I can use `Ctype.equal` for comparing type parameters, but I think this is a simpler way...

This special case needs to be expanded to handle other type descs (including Tarrow). I didn't do the other cases since I wasn't sure if this were the best way to handle the binding of type variables..
I will add a test to exploit that, and handle the other cases.

lib/diff.ml
```diff
@@ -152,32 +174,59 @@ let rec type_item ~typing_env ~name ~reference ~current =
       type_decls ~typing_env ~name ~reference ~current
 
 and type_decls ~typing_env ~name ~reference ~current =
+  let reference, current =
+    if
+      Util.normalized_type_params ~reference:reference.Types.type_params
+        ~current:current.Types.type_params
+    then (reference, current)
+    else Util.normalize_type_decls ~reference ~current
```
NathanReb, line 182: I like this idea because it means we try to preserve the original names as much as possible. I wonder if we couldn't simplify it into:
```suggestion
    Util.normalize_type_decls ~reference ~current
```
by making `normalize` keep shared names and assign new names where they differ between the two declarations. That way we only process the type parameters once and we can even have more sharing than all or nothing.

What do you think?

lib/diff.ml
```diff
@@ -128,8 +131,27 @@ let module_type_fallback ~loc ~typing_env ~name ~reference ~current =
       Some (Module { mname = name; mdiff = Modified Unsupported })
 
 let type_expr typing_env reference current =
-  if Ctype.does_match typing_env reference current then None
-  else Some (Modified { reference; current })
+  let rec same_type_var_names reference current =
```
NathanReb, line 134: This is not required. I think the whole function can instead be rewritten as:
```ocaml
let type_expr ~typing_env ?(ref_params=[]) ?(cur_params=[]) ~reference ~current () =
  if Ctype.is_equal typing_env true (ref_params@[reference]) (cur_params@[current]) then
    None
  else
    Some (Modified {reference; current})
```

There might be some fine tuning to do here, e.g. there might be some extra steps like cloning the type_expr before calling `is_equal` but in the end it should all work with a single call to `Ctype.is_equal`, no manual Tvar matching to do!

lib/diff.ml
```diff
@@ -250,8 +299,13 @@ and modified_variant_type ~typing_env ~ref_constructor_lst ~cur_constructor_lst
   in
   modified_cstrs
 
-and diff_list (diff_one : 'a option -> 'a option -> ('a, 'diff) maybe_changed)
-    (ref : 'a list) (curr : 'a list) : ('a, 'diff) maybe_changed list =
+and diff_list :
+      'a 'diff.
+      ('a option -> 'a option -> ('a, 'diff) maybe_changed) ->
```
NathanReb, line 304: Any particular reason why you removed the arguments labels here?

I think it would be best to keep the labels here!

tests/api-diff/type_manifest_tests.t
```diff
@@ -45,3 +45,4 @@ Run the api-watcher on the two cmi files
   +type t = private int list
   
   [1]
+
```
NathanReb, line 48: Unwanted empty line here I think!

lib/util.ml
```diff
@@ -0,0 +1,163 @@
...
+      (String_map.empty, String_map.empty)
+  in
+  ( normalize_type_decl reference ref_tvmap,
+    normalize_type_decl current cur_tvmap )
+
+let rec normalized_type_params ~reference ~current =
+  match (reference, current) with
+  | [], [] -> true
+  | _ :: _, [] -> true
+  | [], _ :: _ -> true
```
panglesd, line 158:
```suggestion
  | [], _ | _, [] -> true
```

lib/util.ml
panglesd, line 1: Nitpick: I would name this file `normalize.ml`, and remove the `normalize_` prefix on value names. This way, calling them from outside reads: `Normalize.type_decls` (and if we need to normalize other things, `Normalize.values`, etc).

Apart from the naming habits, which you are free to ignore or have your own preferences, I think it is nice to keep the normalizing code free from other logic (it consists of some nicely self-contained 163 lines). `Utils` seems to imply we are going to add more "util" code here...

panglesd (approved):
Looks great! I left some nitpicks.

lib/diff.ml
azzsal: I am not sure if it would look nice when we print the the textual diff. Also there is a problem that the unique names generated could end up colliding with a name of a type parameter that's the same in the two type declaration. 

For example:
```

type ('a, 'b, ..., 't1) = ...

type ('a, 'c, ..., 't1) = ...
```
Though I think it's possible to resolve the name collision by building the maps first, but still having 't2 before 't1 doesn't look that nice to me.

lib/normalize.ml
```diff
@@ -0,0 +1,64 @@
...
+
+let type_params_arity ~reference ~current =
+  if List.length reference = List.length current then (reference, current)
+  else if List.length reference > List.length current then
+    let rest =
+      List.init
+        (List.length reference - List.length current)
+        (fun _ -> create_expr (Tvar None) ~level:0 ~scope:0 ~id:0)
+    in
+    (reference, current @ rest)
```
NathanReb, line 57: This block and the next can be factored out!

lib/normalize.ml
```diff
@@ -0,0 +1,64 @@
...
+  match (reference, current) with
+  | [], _ | _, [] -> true
+  | ref_type_param :: reference', cur_type_param :: current' ->
+      String.equal
+        (get_type_param_name ref_type_param)
+        (get_type_param_name cur_type_param)
+      && is_type_params ~reference:reference' ~current:current'
+
+let type_params_arity ~reference ~current =
+  if List.length reference = List.length current then (reference, current)
```
NathanReb, line 50: `List.length` is a function that will traverse the whole list, I'd suggest
```suggestion
  let ref_len = List.length reference in
  let cur_len = List.lenght current in
  if ref_len = cur_len then (reference, current)
```

tests/api-watch/test_normalize.ml
```diff
@@ -0,0 +1,72 @@
...
+          (Tvar (Some (CCString.of_char (Char.chr (Char.code 'a' + i)))))
+          ~level:0 ~scope:0 ~id:i)
+  in
+  let current =
+    List.init 3 (fun i ->
+        Types.create_expr
+          (Tvar (Some (CCString.of_char (Char.chr (Char.code 'a' + i)))))
+          ~level:0 ~scope:0 ~id:i)
+  in
+  Normalize.is_type_params ~reference ~current
```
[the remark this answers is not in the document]
NathanReb: Without an expect block, nothing is actually tested here. The function can return `true` or `false` without impacting the test output.

lib/normalize.ml
NathanReb: E.g. extract a function with a signature like:
```ocaml
val prepend_tvar_none : type_expr list -> int -> type_expr list
```

NathanReb (commented):
This looks great!

Glad you managed to fix the normalization bug we found by mutating the expression!

lib/diff.ml
NathanReb: Yeah you're right, it would be too much work to have something correct there and probably not worth the pain!

PR #117, opened 2025-01-28
Add type kind to type declaration diff representation

### LINKED ISSUE #114: Represent a change in type_decl_kind when diffing type_declarations
At the moment, when we diff two type_declarations there are two scenarios:
1. Both type_declarations share the same `type_decl_kind` and are either a record or variant type decl: we manually diff the concrete definition to produce a fine grained diff
2. They are not a record or variant type or do not share the same `type_decl_kind`: we represent the whole diff as an atomic modification

What we would like to have is to manually diff the whole `type_decl_kind` field so that:
- If both are record types, we manually diff the fields as we do now
- If both are variants, we manually diff the constructors as we do now
- If both are abstract or open we consider their `type_decl_kind` equal (we can ignore the `Type_abstract` argument).
- If they have different type_decl_kind, we return a `type_decl_kind atomic_modification`

azzsal:
This should resolve #114.

The PR will have the following changes:

- Change the type declaration diff representation to be a record type, the only current field now the `type_kind_mismatch`
- Update the Text_diff module accordingly 
- New cram tests to test the above

### DIFF
--- a/lib/diff.ml
+++ b/lib/diff.ml
@@ -13,17 +13,12 @@ type value = {
 }
 
 type type_ = { tname : string; tdiff : (type_declaration, type_modification) t }
+and type_modification = { type_kind_mismatch : type_kind_mismatch option }
 
-and type_modification =
-  type_kind_mismatch * type_args_mismatch option * type_param list
-
-and type_kind_mismatch = type_kind * type_kind
-and type_kind = Kind_record | Kind_variant | Kind_abstract | Kind_open
-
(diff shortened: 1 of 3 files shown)


azzsal:
I need to add more tests, beside that it's ready for review.

azzsal:
Closing this, I will continue this in #119.

azzsal:
Closing this, I will continue this in https://github.com/ocaml-semver/ocaml-api-watch/pull/119.

Issue #116, opened 2025-01-28
Represent manifest change in type_declaration diff


NathanReb:
We should detect changes in the type manifest when diffing type_declaration and add this information to our fine grained diff representation.

The type manifest is used to declare one's type as an alias of another type. It can be set alongside the concrete declaration. Let me give a few examples. Here are two types with a manifest.

```ocaml
type t = int list
       ^^^^^^^^^^
```
```ocaml
type t = M.t = A of int | B of string
       ^^^^^
```

In the first example the `type_manifest` will be `Some te` where `te` is the `type_expr` for `int list`.

Issue #115, opened 2025-01-28
Represent privacy change


NathanReb:
We should be able to detect if a type changes its privacy or if it remains unchanged by diffing the `type_private` field.

This should add to the diff we compute for the type_decl_kind, meaning a fine grained diff for type_declarations should both have a fine grained diff for the `type_private` and for the `type_decl_kind`.

PR #120, opened 2025-01-30
Add type kind,  privacy and manifest changes to type declaration diff representation

### LINKED ISSUE #114: Represent a change in type_decl_kind when diffing type_declarations

azzsal:
This should resolve #114, #115 and #116.

The PR will have the following changes:

- [x] Update the type declaration diff representation to be a record type, having type_kind, type_privacy and type_manifest diffs
- [x] Update the Text_diff module accordingly
- [x] New cram tests to test the above

### DIFF
--- a/lib/diff.ml
+++ b/lib/diff.ml
@@ -15,20 +15,20 @@ type value = {
 type type_ = { tname : string; tdiff : (type_declaration, type_modification) t }
 
 and type_modification = {
-  type_kind_mismatch : type_kind_mismatch option;
-  type_privacy_mismatch : (Asttypes.private_flag, type_privacy_diff) Either.t;
-  type_manifest_mismatch :
+  type_kind : type_kind option;
+  type_privacy : (Asttypes.private_flag, type_privacy) Either.t;
+  type_manifest :
     (type_expr option, (type_expr, type_expr atomic_modification) t) Either.t;
 }
 
-and type_privacy_diff =
-  | Added_p of Asttypes.private_flag
-  | Removed_p of Asttypes.private_flag
+and type_privacy =
+  | Added_p
+  | Removed_p
 
-and type_kind_mismatch =
-  | Record_mismatch of record_field list
-  | Variant_mismatch of constructor_ list
-  | Atomic_mismatch of type_decl_kind atomic_modification
+and type_kind =
+  | Record_tk of record_field list
+  | Variant_tk of constructor_ list
+  | Atomic_tk of type_decl_kind atomic_modification
 
 and record_field = {
   rname : string;
@@ -47,9 +47,6 @@ and constructor_modification =
 
 and tuple_component =
   (type_expr, (type_expr, type_expr atomic_modification) t) Either.t
-
-and type_param = (type_expr, type_param_diff) Either.t
-and type_param_diff = Added_tp of type_expr | Removed_tp of type_expr
 
 type class_ = {
   cname : string;
@@ -126,6 +123,9 @@ let module_type_fallback ~loc ~typing_env ~name ~reference ~current =
   | exception Includemod.Error _ ->
       Some (Module { mname = name; mdiff = Modified Unsupported })
 
+let same_type_expr typing_env reference current = 
+  Ctype.does_match typing_env reference current
+
 let extract_lbls lbls =
   List.fold_left
     (fun map lbl -> String_map.add (Ident.name lbl.ld_id) lbl map)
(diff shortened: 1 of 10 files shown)


NathanReb (commented):
This is looking good!

The diffing part seems pretty much ready, there are a few places where style can be improved and some simple refactoring can be done to simplify the code a bit but it's pretty much ready!

The textual part is a bit too far off from regular looking OCaml interfaces but I agree it's a bit tedious to basically re-implement the full printing utilities here. As a rule of thumb, I think all information that is not part of the concrete record or variant type definition should be printed on the first line, that includes manifest, privacy and type parameters.
I'm gonna give you a few examples of how we'd like diff to show up for now:
```
-type t = private
+type t =
   {
   ...
+   a : int
   }
```
```
- type t = ..
+ type t =
+  | A
+  | B of int
```
```
- type t =
-  { a : int
-  ; b : string }
+ type t
```

NathanReb (commented):
This is starting to look good! Some of the textual diff output are incorrect but besides that I have only minor suggestions to improve code sharing.

I feel like the textual diff part can be simplified quite a lot but given we are going to change it sooner rather than later, I think we can keep it like this, as long as it produces correct diffs.

lib/text_diff.ml
```diff
@@ -119,24 +126,229 @@ let process_atomic_diff (diff : (_, _ Diff.atomic_modification) Diff.t) name
...
+  let indentation = String.init n (fun _ -> ' ') in
+  indentation ^ s
+
+and process_modified_record_type_diff ~indent_amount diff =
   let changes = process_modified_labels diff in
-  [ indent indent_n (Same "...") ]
-  @ List.map (fun c -> indent indent_n c) changes
+  [ indent indent_amount (Same "{") ]
+  @ [ indent (indent_amount + 2) (Same "...") ]
+  @ List.map (fun c -> indent (indent_amount + 2) c) changes
```
NathanReb, line 350: You don't need to construct a singleton list and concatenate here, you can simply append:
```suggestion
  (indent indent_amount (Same "{"))
  :: (indent (indent_amount + 2) (Same "..."))
  :: (List.map (fun c -> indent (indent_amount + 2) c) changes)
```

lib/text_diff.ml
```diff
@@ -119,24 +126,229 @@ let process_atomic_diff (diff : (_, _ Diff.atomic_modification) Diff.t) name
...
+  | Types.Type_variant (cstr_lst, _) ->
+      Some
+        (List.map
+           (fun s -> indent_s 1 s)
+           (List.concat_map (fun cstr -> cstr_to_lines cstr) cstr_lst))
+  | Types.Type_abstract _ -> None
+  | Types.Type_open -> Some [ indent_s 1 ".." ]
+
+and indent_s n s =
+  let indentation = String.init n (fun _ -> ' ') in
```
NathanReb, line 343:
```suggestion
  let indentation = String.make n ' ' in
```

NathanReb (commented):
Looks great! I have a couple small suggestions but then it's good to go!

tests/api-diff/type_tests.t
```diff
@@ -70,5 +70,6 @@ Run api-watcher on the two cmi files, there should be a difference
   diff module Modify_type:
   -type t = int
   +type t = float
+
```
panglesd, line 73: Any idea why this new line is added (it seems there is a lot of new lines at the end of each diff)

panglesd (commented):
Minor comments!

lib/text_diff.ml
```diff
@@ -119,24 +126,229 @@ let process_atomic_diff (diff : (_, _ Diff.atomic_modification) Diff.t) name
...
+  | ( Either.Left private_flag,
+      Either.Left te_opt,
+      Either.Right
+        (Atomic_tk
+          { reference = Types.Type_open; current = Types.Type_record (_, _) }) )
+  | ( Either.Left private_flag,
+      Either.Left te_opt,
+      Either.Right
+        (Atomic_tk
+          { reference = Types.Type_open; current = Types.Type_variant (_, _) })
```
panglesd, line 215: Also, I'm pretty sure this pattern-match can be made more explicit: you are only matching on a specific part, all the rest is duplicated. You can do something like this:

```ocaml
 | ( Either.Left private_flag,
      Either.Left te_opt,
      Either.Right
        (Atomic_tk
          (CASE1 | CASE2 | CASE3) ->
```

tests/api-diff/type_tests.t
azzsal: Yes. These are only there when both type_kind aren't changed and they are abstract, in that case there we print a newline for the unchanged type_kind for context.

lib/diff.ml
```diff
@@ -14,10 +14,19 @@ type value = {
 
 type type_ = { tname : string; tdiff : (type_declaration, type_modification) t }
 
-and type_modification =
-  | Record_diff of record_field list
-  | Variant_diff of constructor_ list
-  | Atomic of type_declaration atomic_modification
+and type_modification = {
+  type_kind : (Types.type_decl_kind, type_kind) Either.t;
```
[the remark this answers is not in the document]
azzsal: Should I replace the all the `Either.t` with something like this:
```
type type_kind_change =
  | Same of type_decl_kind
  | Diff of type_kind
```

tests/api-diff/type_tests.t
panglesd: But is there any explanation why we would want a blank line there? I think it would make more sense not to print an empty line in this case!

lib/text_diff.ml
```diff
@@ -119,24 +126,161 @@ let process_atomic_diff (diff : (_, _ Diff.atomic_modification) Diff.t) name
...
+        (Atomic_tk
+          ( { reference = Type_record (_, _); current = Type_record (_, _) }
+          | { reference = Type_record (_, _); current = Type_variant (_, _) }
+          | { reference = Type_record (_, _); current = Type_open }
+          | { reference = Type_variant (_, _); current = Type_record (_, _) }
+          | { reference = Type_variant (_, _); current = Type_variant (_, _) }
+          | { reference = Type_variant (_, _); current = Type_open }
+          | { reference = Type_open; current = Type_record (_, _) }
+          | { reference = Type_open; current = Type_variant (_, _) }
+          | { reference = Type_open; current = Type_open } )) ) ->
```
panglesd, line 173: Now that I see it in a short version, I think you can still make it much more readable with:

```ocaml
[...]
   Atomic_tk { reference = (CASE1 | CASE2 | ...); current = (CASE1 | CASE2 | ... }
```

panglesd: Also, patterns like `(_, _)` can be replaced simply by `_`)

tests/api-diff/type_tests.t
azzsal: I totally agree. But here we have a Same hunk, which has an empty string underneath, to get rid of this newline we need to change the representation a bit, to have the Same hunk hold either a string option or a sting list. I am not sure if it's worth the trouble since we'll change the representation to have inline diffs very soon.

NathanReb: Context lines will still be there even with inline change. Why not remove the `Same` hunk altogether here?

NathanReb (approved):
Looks good to me! If @panglesd agrees we can merge!

PR #126, opened 2025-02-12
Improve textual diff representation

### LINKED ISSUE #122: More readable types in `diff.mli`
The types representing diffed AST has became quite hard to read.

For instance:

```ocaml
and type_modification = {
  type_kind : (Types.type_decl_kind, type_kind) maybe_changed;
  type_privacy : (Asttypes.private_flag, type_privacy) maybe_changed;
  type_manifest :
    ( Types.type_expr option,
      (Types.type_expr, Types.type_expr atomic_modification) t )
    maybe_changed;
  type_params : (Types.type_expr list, type_param list) maybe_changed;
}
```

what is exactly the type for the `type_manifest` field? That's not obvious at first. It is representing a difference between two `Types.type_expr option`:
- If there is no difference, we get the common `Types.type_expr option` (in the `Same` constructor)
- If there is a difference:
  - Either the two values are `Some _` and in this case we represent the diff with `Different (Modified _)` and an atomic `type_expr` modification
  - Or they are `None` and `Some _` and we represent that with `Different Added/Removed` and the `type_expr` value.

So, the implementation is perfectly right, the issue is only about readability! It would be nice to have a common pattern and type aliases to represent various situations:
- A diff between two options
- A diff in two "string maps" (such as a map from field name to field, from signature item name to signature item, ...)
- A diff between two values of a variant type

Moreover, the type originally named `t` which was representing an addition/removal/modification should be renamed to something more explicit, now that we are representing more than just that.

azzsal:
This PR has multiple changes:
- Make the structural diff representation more readable by introducing some type aliases for common diff types
- Improve the structural diff representation to store context
- Update the textual diff representation to be more readable by highlighting changes in a line

### DIFF
--- a/lib/diff.ml
+++ b/lib/diff.ml
@@ -1,79 +1,52 @@
 open Types
+open Stddiff
 
-type ('item, 'diff) t = Added of 'item | Removed of 'item | Modified of 'diff
-
-type 'a atomic_modification = { reference : 'a; current : 'a }
-(** The simplest diff representation for the modification of a value of type 'a.
-    [reference] is the value before and [current] is the value after the change
-    occured. Use this type when there is no better representation available. *)
-
-type value = {
-  vname : string;
-  vdiff : (value_description, value_description atomic_modification) t;
-}
-
-type type_ = { tname : string; tdiff : (type_declaration, type_modification) t }
-
-and type_modification = {
-  type_kind : (Types.type_decl_kind, type_kind) maybe_changed;
+type type_modification = {
+  type_kind : (type_decl_kind, type_kind) atomic_variant_maybe_changed;
   type_privacy : (Asttypes.private_flag, type_privacy) maybe_changed;
-  type_manifest :
-    ( type_expr option,
-      (type_expr, type_expr atomic_modification) t )
-    maybe_changed;
-  type_params : (type_expr list, type_param list) maybe_changed;
+  type_manifest : type_expr atomic_option;
+  type_params : (type_expr, type_param) list_;
 }
-
-and ('same, 'different) maybe_changed =
-  | Same of 'same
-  | Different of 'different
-
-and type_param = (type_expr, type_param_diff) maybe_changed
-and type_param_diff = Added_tp of type_expr | Removed_tp of type_expr
-and type_privacy = Added_p | Removed_p
 
 and type_kind =
-  | Record_tk of record_field list
-  | Variant_tk of constructor_ list
-  | Atomic_tk of type_decl_kind atomic_modification
+  | Record_tk of (label_declaration, label) map
+  | Variant_tk of
+      ( constructor_declaration,
+        (constructor_arguments, cstr_args) atomic_variant )
+      map
 
-and record_field = {
-  rname : string;
-  rdiff : (label_declaration, label_declaration atomic_modification) t;
+and label = {
+  label_type : type_expr maybe_changed_atomic;
+  label_mutable : (Asttypes.mutable_flag, field_mutability) maybe_changed;
 }
 
-and constructor_ = {
-  csname : string;
-  csdiff : (constructor_declaration, constructor_modification) t;
+and field_mutability = Added_m | Removed_m
+
+and cstr_args =
+  | Record_cstr of (label_declaration, label) map
+  | Tuple_cstr of type_expr maybe_changed_atomic_entry list
+
+and type_privacy = Added_p | Removed_p
+and type_param = (type_expr, type_param_diff) maybe_changed
+and type_param_diff = Added_tp of type_expr | Removed_tp of type_expr
+
+type type_ = {
+  tname : string;
+  tdiff : (type_declaration, type_modification) entry;
 }
 
-and constructor_modification =
-  | Record_c of record_field list
-  | Tuple_c of tuple_component list
-  | Atomic_c of constructor_declaration atomic_modification
-
-and tuple_component =
-  (type_expr, (type_expr, type_expr atomic_modification) t) maybe_changed
-
-type class_ = {
-  cname : string;
-  cdiff : (class_declaration, class_declaration atomic_modification) t;
-}
-
-type cltype = {
-  ctname : string;
-  ctdiff :
-    (class_type_declaration, class_type_declaration atomic_modification) t;
-}
+type value = { vname : string; vdiff : value_description atomic_entry }
+type class_ = { cname : string; cdiff : class_declaration atomic_entry }
+type cltype = { ctname : string; ctdiff : class_type_declaration atomic_entry }
 
 type module_ = {
   mname : string;
-  mdiff : (module_declaration, signature_modification) t;
+  mdiff : (module_declaration, signature_modification) entry;
 }
 
 and modtype = {
   mtname : string;
-  mtdiff : (modtype_declaration, signature_modification) t;
+  mtdiff : (modtype_declaration, signature_modification) entry;
 }
 
 and signature_modification = Unsupported | Supported of sig_item list
@@ -117,6 +90,41 @@ let extract_items items =
       | _ -> tbl)
     Sig_item_map.empty items
 
+and diff_map ~(diff_one : 'a -> 'a -> 'diff option) ~(ref_map : 'a String_map.t)
+    ~(cur_map : 'a String_map.t) : ('a, 'diff) map =
+  let same_seq, changed_seq =
+    String_map.merge
+      (fun _ ref_opt cur_opt ->
+        match (ref_opt, cur_opt) with
+        | None, None -> None
+        | Some ref, None -> Some (Changed (Removed ref))
+        | None, Some cur -> Some (Changed (Added cur))
+        | Some ref, Some cur -> (
+            match diff_one ref cur with
+            | None -> Some (Same ref)
+            | Some diff -> Some (Changed (Modified diff))))
+      ref_map cur_map
+    |> String_map.to_seq
+    |> Seq.partition_map (fun (name, i) ->
+           match i with
+           | Same i -> Either.Left (name, i)
+           | Changed i -> Either.Right (name, i))
+  in
+  {
+    same_map = String_map.of_seq same_seq;
+    changed_map = String_map.of_seq changed_seq;
+  }
+
+let extract_lbls lbls =
+  List.fold_left
+    (fun map lbl -> String_map.add (Ident.name lbl.ld_id) lbl map)
+    String_map.empty lbls
+
+let extract_cstrs cstrs =
+  List.fold_left
+    (fun map cstr -> String_map.add (Ident.name cstr.cd_id) cstr map)
+    String_map.empty cstrs
+
 let module_type_fallback ~loc ~typing_env ~name ~reference ~current =
   let modtype_coercion1 () =
     Includemod.modtypes ~loc typing_env ~mark:Mark_both reference current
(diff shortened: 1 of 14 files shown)


azzsal:
@panglesd I have added some type aliases as you suggested to have shorter diff types.
Is there anything you like me to add/change before I stream the changes to the Diff and Text_diff modules ?

panglesd (approved):
I added a few suggestion but all the aliases that were added seem good to me and I think makes it a bit more readable. Thanks!

lib/diff.mli
```diff
@@ -1,100 +1,82 @@
...
 
-type 'same atomic_option =
-  ('same, 'same atomic_modification) option_
+type ('same, 'diff) option_ = ('same option, ('same, 'diff) entry) maybe_changed
+type 'same atomic_option = ('same, 'same atomic_modification) option_
 
 type ('same, 'diff) atomic_variant =
   ('same, ('same atomic_maybe_changed, 'diff) maybe_changed) maybe_changed
 
 type ('same, 'diff) list_ = ('same list, 'diff list) maybe_changed
```
panglesd, line 28: I think the `list_` should include more than this: it's a list of things than may or may not have changed, and if they changed, it can be an addition, removal or modification.

I believe this will allow to remove the confusing `atomic_maybe_changed_entry` (which is hard to understand what it's for from the name) and hopefully simplify the usage of `list_`.

lib/diff.mli
```diff
@@ -1,100 +1,82 @@
...
-and modtype = {
-  mtname : string;
-  mtdiff : (Types.modtype_declaration, signature_modification) t;
+  mdiff : (Types.module_declaration, signature_modification) entry;
 }
 
-and signature_modification = Unsupported | Supported of sig_item list
+and signature_modification =
+  | Unsupported
+  | Supported of (Types.signature_item, sig_item) map
```
panglesd, line 71: Unfortunately this has either to be several maps, or not to be "string_maps": the namespace for values and types are not the same, so as it is written there could be clashes.

panglesd (commented):
I left a few suggestion, and there is a small problem with namespaces for signature!

Let me know if propagating the type changes cause any problem. I can help with that!

lib/diff.mli
```diff
@@ -1,29 +1,33 @@
 type ('item, 'diff) entry =
   | Added of 'item
   | Removed of 'item
-  | Modified of 'diff
+  | Modified of 'diff  (** Represent a change of an entry in a collection *)
```
panglesd, line 4: Comments that start with two `*` are special comments which represent documentation strings.
They are attached to a specific place in the AST, depending on where they are placed. If they are "touching" an item, they are attached to it:

```ocaml
type t = int
(** I'm a docstring attached to t *)

(** I'm a docstring attached to u *)
type u = int
```
You can also attach docstring to variant declaration:

```ocaml
(** I'm attached to t *)
type t =
  | A of int (** I'm attached to A *)
  | B of int
```

You can see [here](https://ocaml.org/manual/5.3/api/compilerlibref/Parsetree.html#1_Extensionpoints) how comments attached to items and constructors are rendered differently!

However, to what should this comment be associated?

```ocaml
type t =
  | A of int
  | B of int
(** Am I associated to B or to t ? *)
```

The comment above is associated to `B`!

All of this just to say that you should put the docstring just before `type ... entry` so that it is associated to the type, and not to the last constructor.

lib/diff.mli
```diff
@@ -35,54 +39,50 @@ and type_modification = {
...
-  | Variant_tk of
-      ( Types.constructor_declaration,
-        (Types.constructor_declaration, cstr_args) atomic_variant )
-      map
-
-and record_field = {
-  type_expr : Types.type_expr atomic_maybe_changed;
-  mutable_flag : (Asttypes.mutable_flag, field_mutability) maybe_changed;
+  | Record_tk of (Types.label_declaration, label) map
+  | Variant_tk of (Types.constructor_declaration, cstr_args) map
```
panglesd, line 43: Why not use the `variant` label here, instead of having a `Atomic_cstr` in `cstr_args`?

lib/diff.mli
```diff
@@ -1,29 +1,33 @@
...
-type 'same atomic_maybe_changed_entry =
+type 'same maybe_changed_atomic_entry =
   ('same, 'same atomic_entry) maybe_changed
 
 type ('same, 'diff) option_ = ('same option, ('same, 'diff) entry) maybe_changed
 type 'same atomic_option = ('same, 'same atomic_modification) option_
 
 type ('same, 'diff) atomic_variant =
-  ('same, ('same atomic_maybe_changed, 'diff) maybe_changed) maybe_changed
+  ('same, ('same atomic_modification, 'diff) Either.t) maybe_changed
```
panglesd, line 30: Why don't use use a `maybe_changed` here?
`Same` would mean "same constructor with different argument values" and `Changed` would mean "different constructors".

Seeing `Left` and `Right` in the middle of code will not be self-explanatory on what they represent!

NathanReb: I agree the `Either.t` is not the best suited type here but I don't think we should hide a diff in the `Same of x` argument of a `maybe_changed` as this is highly inconsistent with what it represent and likely going to confuse users.

lib/diff.ml
```diff
@@ -29,7 +29,7 @@ type ('same, 'diff) option_ = ('same option, ('same, 'diff) entry) maybe_changed
 type 'same atomic_option = ('same, 'same atomic_modification) option_
 
 type ('same, 'diff) atomic_variant =
-  ('same, ('same atomic_modification, 'diff) Either.t) maybe_changed
+  ('same, ('same atomic_modification, 'diff) maybe_changed) maybe_changed
```
panglesd, line 32: My suggestion was "Change means changed variant and Same means same variant" so you would need to invert the type params here:
```ocaml
type ('same, 'diff) atomic_variant =
  ('same, ('diff, 'same atomic_modification) maybe_changed) maybe_changed
```
however, I think Nathan suggested to define another variant here:
```
type ('same_variant, 'different_variant) variant_change =
  | Same_variant of 'same_variant
  | Different_variant of 'different_variant
```

(can you confirm that was your suggestion @NathanReb ?)

NathanReb: Yes I think that would be better than to hide a diff in the "no diff" side of a `maybe_changed`!

NathanReb:
I haven't gone through the whole thing yet but it might be good to extract the diff types for common OCaml types into a separate module!

lib/diff.mli
azzsal: I agree. But that's not always the case with a list diff type. For example with type parameters, it's either same or changed, and the change could be either an added or removed type parameter (we don't have modified here). so I think we have to specify the kind of diff we have as a parameter for type `list_`.

I was wondering if it would make up for a good use case of extensible variants here ?
Like having the top entry type be an extensible type as this: `type ('item, 'diff) entry = ..`
and then extending it as needed with different constructors.
Like this for type paramters: 
```ml 
type ('item, `diff) entry += Added_tp
type ('item, 'diff) entry += Removed_tp
```

lib/text_diff.ml
```diff
@@ -1,24 +1,56 @@
-type conflict2 = { orig : string list; new_ : string list }
-type hunk = Change of conflict2 | Same of string
-type t = hunk list String_map.t
-type printer = { same : string Fmt.t; diff : conflict2 Fmt.t }
+type inline_conflict = { iorig : string option; inew : string option }
+type inline_hunk = Icommon of string option | Iconflict of inline_conflict
```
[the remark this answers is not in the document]
NathanReb: I'd suggest removing the option here and in the various `process_` functions instead to wrap the `inline_hunk` in options before doing a final `List.filter_map id`.

This is an implementation detail leaking into our public API.

lib/text_diff.ml
```diff
@@ -134,272 +149,350 @@ let rec process_type_diff (type_diff : Diff.type_) =
...
+  let space =
+    match params_hunks with _ :: _ -> Icommon (Some " ") | _ -> Icommon None
+  in
+  (space :: open_paren :: params_hunks) @ [ close_paren ]
+
+and process_equal_sign_diff type_privacy_diff type_manifest_diff type_kind_diff
+    =
+  let open Types in
+  let open Asttypes in
+  match (type_privacy_diff, type_manifest_diff, type_kind_diff) with
```
NathanReb, line 289: Couldn't this be simplified? I feel like the `type_manifest` should be dealt with separately since here we're deciding whether to add an `=` sign after the manifest.

It also seems that the privacy does not strictly matter does it?

lib/text_diff.ml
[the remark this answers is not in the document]
azzsal: I made the `process_` functions return a list, an empty list would indicate that there are no inline hunks. Sorry about any inconsistencies.

lib/text_diff.ml
azzsal: I think we do have to check for the `type_manifest` and `type_privacy` to decide whether to add an equal sign or not,  These come after the equal sign.

lib/diff.ml
```diff
@@ -1,79 +1,100 @@
...
-  | Atomic_tk of type_decl_kind atomic_modification
+  type ('same, 'diff) option_ =
+    ('same option, ('same, 'diff) entry) maybe_changed
+
+  type 'same atomic_option = ('same, 'same atomic_modification) option_
 
-and record_field = {
-  rname : string;
-  rdiff : (label_declaration, label_declaration atomic_modification) t;
+  type ('same, 'diff) variant =
```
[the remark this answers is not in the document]
azzsal: It was one type before like this I believe: 

```
type ('same, 'diff) atomic_variant =
  ('same, ('same atomic_modification, 'diff) maybe_changed) maybe_changed
```

The problem is the diff is hidden in the maybe changed, should I revert back to that ?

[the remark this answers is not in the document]
NathanReb: No, what I suggest here is that there is no need to wrap the `type_kind` diff or the `constructor_argument` diff inside another layer.

E.g. reinstate the fact that when they both value have a different constructor, it's encoded into the diff type directly rather than via a generic variant_type_diff wrapper, i.e. instead of:
```ocaml
Same_variant (Record_tk ...)
```
or
```ocaml
Different_variant {orig = Type_record ...; new_ = Type_variant ...}
```
but rather define the following `Diff.type_kind`
```ocaml
type_kind =
  | Record_tk of (Types.label_declaration, label) Stddiff.map
  | Variant_tk of (Types.constructor_declaration, constructor_diff) Stddiff.map
  | No_sharing of {orig: type_decl_kind; new_ = type_decl_kind}
```

We take away layers of types and type parameters that hurt the comprehension of the what the types actually are and how to construct values for those types. I think there is no reason to try and generalize a type here, given we always have to have at least one specialized diff type for a variant type.

lib/text_diff.ml
NathanReb: I don't think so. There can be up to two equal signs in the header, one for the concrete definition and one for the manifest. The manifest always come first so it should be dealt with separately, i.e. if there is no manifest, we don't have a manifest hunk, if there is a manifest, the manifest hunk contains the leading equal. Then, based on what the concrete declaration is, we decide whether there is an equal sign. Let me illustrate with an example:
```ocaml
type 'a t = 'a list =
  | Nil
  | Cons of 'a * 'a list
```
The header here can be split into the following hunks:
```ocaml
[Icommon "type"; Icommon "'a"; Icommon"t"; Icommon "= 'a list"; Icommon "="]
```
I omitted the spaces and wrote everything into an `Icommon` for the sake of simplicity but you get the idea. The last hunk depends solely on the type_kind (it has to be record, variant or open). The second to last hunk depends solely on the manifest value.

Now if we were to make this type private, it would look like this:
```ocaml
type 'a t = 'a list = private
  | Nil
  | Cons of 'a * 'a list
```
The private comes after the concrete definition `=`. A type without a concrete definition cannot be private so there's no need to bother with this but to err on the safe side, we can generate the privacy hunk based on the value of the private flag and concrete definition. So we should generate something like:
```ocaml
[Icommon "type"; Icommon "'a"; Icommon"t"; Icommon "= 'a list"; Icommon "="; Icommon "private"]
```

NathanReb (commented):
Looks good! As discussed offline, I think any further improvement on the diff -> textual diff conversion for type_declaration diff will require changing the representation, I'll open a separate issue about that!

PR #128, opened 2025-02-18
Fix loader with paths to module types

### LINKED ISSUE #121: Loading sometimes fail
Here is a minimal (?) reproduction:

```
  $ mkdir repro
  $ cd repro

  $ cat > dune-project <<EOF
  > (lang dune 2.9)
  > EOF

  $ cat > dune <<EOF
  > (library
  >  (name base))
  > EOF

  $ cat > bytes_intf.ml <<EOF
  > module type X = sig end
  > module type Y = sig
  >  module Z : X
  > end
  > EOF

  $ cat > bytes.ml <<EOF
  > module Z = struct end
  > EOF

  $ cat > bytes.mli <<EOF
  > include Bytes_intf.Y
  > EOF

  $ dune build

  $ api-diff --main-module base _build/default/.base.objs/byte _build/default/.base.objs/byte
  api-diff: Could not find module X in Base__Bytes_intf
  [123]
```

I found that while trying to use the tool to diff two different versions of base.

panglesd:
Fix #121 (with a simplified minimal reproduction in the test).

Consider the following signature items:
```
module A = A.B.X
module B : A.B.Y
```

Here, the module type of `A` is the module type of `A.B.Y` (which is a path to a module). This is represented in `Types` as `Mty_alias path`.
The module type of `B` is `A.B.Y` (which is a path to a module _type_). This is represented in `Types` as `Mty_ident path`.

Before this PR, the two cases were treated equally, as if they were path to modules.

We now distinguish the two cases.

(This PR does not address the fact that we do not expand module types during loading! But we are not diffing them anyway as of today.)

### DIFF
--- a/lib/library.ml
+++ b/lib/library.ml
@@ -1,3 +1,6 @@
+let ( let> ) x f =
+  match x with Ok None -> Ok None | Ok (Some x) -> f x | Error x -> Error x
+
 let mod_name file =
   String.capitalize_ascii Filename.(remove_extension (basename file))
 
@@ -31,11 +34,12 @@ let load_cmi file_path =
     Ok (cmi_infos.cmi_sign, cmi_infos.cmi_name)
   with e -> Error (Printexc.to_string e)
 
-(* Attach a module name to its various representations, e.g. a [signature] or a
+type 'a named = { name : string; value : 'a }
+(** Attach a module name to its various representations, e.g. a [signature] or a
    [module_type].
    Mostly used to report lookup failures. *)
-type 'a named = { name : string; value : 'a }
 
+(** A module for "flat path", paths without a functor application in them. *)
 module Flat_path = struct
   type component = Id of Ident.t | Comp of string
   type t = component list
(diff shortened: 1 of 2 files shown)


lib/library.ml
```diff
@@ -83,101 +86,118 @@ let find_module modname sig_ =
...
-        | Some flat_mty_path ->
-            find_module_in_lib ~library_modules flat_mty_path
+  let* sig_ = sig_of_module_type ~library_modules mty.value in
+  match sig_ with
+  | None ->
+      let res =
+        match mty.value with
+        | Mty_alias mty_path | Mty_ident mty_path ->
+            let expanded_path = path_in_module ~module_path:mty_path path in
+            rewrite_mty_path mty.value expanded_path
```
panglesd, line 125: Maybe a comment on why we need this rewriting is worth it!

NathanReb (approved):
Looks good to me!

You raise valid concerns regarding some bits of the loading code, I'll add test and specify those.

NathanReb:
Don't forget to update the changelog!

PR #131, opened 2025-02-21
Add word based textual diff output

### LINKED ISSUE #130: Inline textual diffs
As with #126 we should have line-based textual diffs of type declarations. 
The exact changes in a type declarations should be highlighted in a line, and common parts repeated on the two lines.

Another way to display is to have a word-based textual diff output where both the exact changes and the common parts are displayed on a single where the changes are wrapped in `[-removed-]` and `{added}`, and common parts are displayed only once. This could be useful in cases where there is only and a small change on a line.

Say we have the following "long" record type:

```ml
type t = { a : int; b : float; c : char; d : string; e : unit }
```

and only one field type changed:

```ml
type t = { a : int;  b : float; c : char; d : string; e : int }
```

with line-based text diff output:

```
-type t = { a : int; b : float; c : char; d : string; e : unit }
+type t = { a : int;  b : float; c : char; d : string; e : int }
```

It can sometimes be easier to read word-based diff output like the following:

```
type t = { a : int; b : float; c : char; d : string; e : [-unit-] {+int+}; }
```

azzsal:
This should resolve #130.

The following changes will be made:

- [x] Adding --word-diff command line flag.
- [x] Adding word based diff pretty printers.

### DIFF
--- a/bin/api_diff.ml
+++ b/bin/api_diff.ml
@@ -4,7 +4,7 @@ type mode = Unwrapped | Wrapped of string | Cmi
 type word = Plain | Color
 type display = Line | Word of word
 
-let display_mode word : (display, string) result =
+let display_mode word =
   match word with
   | Some "plain" -> Ok (Word Plain)
   | Some "color" -> Ok (Word Color)
@@ -53,17 +53,13 @@ let mode ~reference ~current ~main_module ~unwrapped =
         "Either --main-module or --unwrapped must be provided when diffing \
          entire libraries."
 
-let print_diff diff word_diff =
-  let open CCResult.Infix in
-  let text_diff = Api_watch.Text_diff.from_diff diff in
-  let* display_mode = display_mode word_diff in
-  (match display_mode with
+let print_diff text_diff display_mode =
+  match display_mode with
   | Line -> Api_watch.Text_diff.With_colors.pp Format.std_formatter text_diff
   | Word Plain ->
       Api_watch.Text_diff.Word.pp ~mode:`Plain Format.std_formatter text_diff
   | Word Color ->
-      Api_watch.Text_diff.Word.pp ~mode:`Color Format.std_formatter text_diff);
-  Ok ()
+      Api_watch.Text_diff.Word.pp ~mode:`Color Format.std_formatter text_diff
 
 let run (`Word_diff word_diff) (`Main_module main_module)
     (`Unwrapped_library unwrapped) (`Ref_cmi reference) (`Current_cmi current) =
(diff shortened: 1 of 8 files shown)


bin/api_diff.ml
```diff
@@ -75,12 +75,24 @@ let run (`Main_module main_module) (`Unwrapped_library unwrapped)
...
     diff_map;
   if has_changes then Ok 1 else Ok 0
 
 let named f = Cmdliner.Term.(app (const f))
 
+let word_diff =
+  let doc =
+    "Show changes in a signature item inline, where a removed part of an item \
+     is wrapped in\n\
+    \      $(b,[-removed-]) and an added one is wrapped in $(b,{+added+})"
```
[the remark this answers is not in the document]
NathanReb: I think it'd be nice to allow both the `plain` and `color` versions!

lib/text_diff.ml
```diff
@@ -1,4 +1,8 @@
-type inline_conflict = { iorig : string option; inew : string option }
+type inline_conflict = {
+  iorig : string option * string option;
```
NathanReb, line 2: I don't understand this new representation, what are each element of the pair?

NathanReb (commented):
I don't think adding a different printer for an inline word diff format should require any change in how we represent and/or produce the textual diff abstraction.

This PR should limit itself to a different way of printing the same diff abstraction.

If this is just to avoid having a few spaces included in conflicts, I think we can both do it separately and find a better representation for it!

lib/text_diff.ml
azzsal: Sorry, this should've been a record. I think I will remove it altogether and restore the old representation.
I agree changing the representation to accommodate for "punctuation" (spaces, commas, etc.) should be done in a separate PR.

bin/api_diff.ml
```diff
@@ -40,8 +53,20 @@ let mode ~reference ~current ~main_module ~unwrapped =
         "Either --main-module or --unwrapped must be provided when diffing \
          entire libraries."
 
-let run (`Main_module main_module) (`Unwrapped_library unwrapped)
-    (`Ref_cmi reference) (`Current_cmi current) =
+let print_diff diff word_diff =
+  let open CCResult.Infix in
+  let text_diff = Api_watch.Text_diff.from_diff diff in
+  let* display_mode = display_mode word_diff in
```
NathanReb, line 59: This can be moved outside of this function as it only needs to be computed once for the whole program. You can make `print_diff` take the `display_mode` directly.

This allow to use `List.iter` too, without poluting it with the `result`!

NathanReb (commented):
This looks great! Thanks!

I only have one minor suggestion to simplify the handling of the display mode but other than that, it's good to go!

PR #133, opened 2025-03-02
Introduce an intermediate representation for type declarations

### LINKED ISSUE #132: Improve type_declarations and associated diff representation
The compiler's representation for type declaration is a large record field, defined as follows in `typing/types.mli`:
```ocaml
type type_declaration =
  { type_params: type_expr list;
    type_arity: int;
    type_kind: type_decl_kind;
    type_private: private_flag;
    type_manifest: type_expr option;
    type_variance: Variance.t list;
    (* covariant, contravariant, weakly contravariant, injective *)
    type_separability: Separability.t list;
    type_is_newtype: bool;
    type_expansion_scope: int;
    type_loc: Location.t;
    type_attributes: Parsetree.attributes;
    type_immediate: Type_immediacy.t;
    type_unboxed_default: bool;
    (* true if the unboxed-ness of this type was chosen by a compiler flag *)
    type_uid: Uid.t;
  }

and type_decl_kind = (label_declaration, constructor_declaration) type_kind

and ('lbl, 'cstr) type_kind =
    Type_abstract of type_origin
  | Type_record of 'lbl list  * record_representation
  | Type_variant of 'cstr list * variant_representation
  | Type_open
```

Interpreting what the type declaration actually represents requires interpreting several of these fields at once and how to interpret/whether to interpret a field at all is dependent on values of some other fields.

Additionally, a lot of those fields are irrelevant for diffing a library's interface.

I suggest we use our own representation for type_declarations.

## Our type_declaration repr

We'll need to convert from the compiler's representation to our own before diffing, and adapt our representation for type_declaration diff accordingly.

My initial suggestion would be something along those lines:
```ocaml
type type_declaration =
  { type_params : type_expr list
   ; type_kind : type_decl_kind
  }

type type_decl_kind =
  | Type_abstract
  | Type_alias of {manifest: type_expr; privacy: private_flag}
  | Type_concrete of {manifest: type_expr option; privacy: private_flag;  definition: concrete_definition}

type concrete_definition =
  | Concrete_open
  | Concrete_record of label_declaration list
  | Concrete_variant of constructor_declaration list
```

I think this would help interpreting the type_declarations and associated diffs for our users and woud greatly simplify the conversion from our structural diff to our internal textual diff representation.

azzsal:
This should resolve #132.

The IR is inspired by/adapted from [odoc internal representation](https://ocaml.github.io/odoc/odoc/src/odoc.model/lang.ml.html#module-TypeDecl). 
We will gradually migrate from the compiler representation to this IR. This PR starts with type declarations.

### DIFF
--- a/lib/convert.ml
+++ b/lib/convert.ml
@@ -1,0 +1,56 @@
+module T = Types
+module TD = Intermed.TypeDecl
+
+let bool_of_private = function
+  | Asttypes.Private -> true
+  | Asttypes.Public -> false
+
+let bool_of_mutable = function
+  | Asttypes.Mutable -> true
+  | Asttypes.Immutable -> false
+
+let field lbl =
+  {
+    TD.Field.id = lbl.T.ld_id;
+    mutable_ = bool_of_mutable lbl.ld_mutable;
+    type_ = lbl.ld_type;
+  }
+
+let cstr_args cd_args =
+  match cd_args with
+  | T.Cstr_tuple type_exprs -> TD.Constructor.Tuple type_exprs
+  | T.Cstr_record lbls -> Record (List.map field lbls)
+
+let cstr cd = { TD.Constructor.id = cd.T.cd_id; args = cstr_args cd.cd_args }
+and param p = { TD.type_expr = p }
+
+let type_declaration ~src =
+  let params = src.T.type_params in
+  let kind =
+    match (src.type_manifest, src.type_kind) with
+    | None, T.Type_abstract _ -> TD.Kind.Abstract
+    | Some manifest, Type_abstract _ ->
+        Alias { manifest; private_ = bool_of_private src.type_private }
+    | manifest, Type_open ->
+        Concrete
+          {
+            manifest;
+            private_ = bool_of_private src.type_private;
+            definition = Open;
+          }
+    | manifest, Type_record (lbls, _) ->
+        Concrete
+          {
+            manifest;
+            private_ = bool_of_private src.type_private;
+            definition = Record (List.map field lbls);
+          }
+    | manifest, Type_variant (cstrs, _) ->
+        Concrete
+          {
+            manifest;
+            private_ = bool_of_private src.type_private;
+            definition = Variant (List.map cstr cstrs);
+          }
+  in
+  { TD.params = List.map param params; kind }
--- a/lib/intermed.ml
+++ b/lib/intermed.ml
@@ -1,0 +1,31 @@
+module rec TypeDecl : sig
+  module Field : sig
+    type t = { id : Ident.t; mutable_ : bool; type_ : Types.type_expr }
+  end
+
+  module Constructor : sig
+    type args = Tuple of Types.type_expr list | Record of Field.t list
+    type t = { id : Ident.t; args : args }
+  end
+
+  type param = { type_expr : Types.type_expr }
+
+  module Kind : sig
+    type definition =
+      | Open
+      | Record of Field.t list
+      | Variant of Constructor.t list
+
+    type t =
+      | Abstract
+      | Alias of { manifest : Types.type_expr; private_ : bool }
+      | Concrete of {
+          manifest : Types.type_expr option;
+          private_ : bool;
+          definition : definition;
+        }
+  end
+
+  type t = { params : param list; kind : Kind.t }
+end =
+  TypeDecl
--- a/lib/convert.mli
+++ b/lib/convert.mli
@@ -1,0 +1,5 @@
+(** Converts from compiler representation of different items (type_declarations,
+    value_description, etc.) to an internal intermediate representation used when diffing.
+*)
+
+val type_declaration : src:Types.type_declaration -> Intermed.TypeDecl.t


lib/intermed.ml
```diff
@@ -0,0 +1,32 @@
...
+  module Field : sig
+    type t = { name : string; mutable_ : bool; type_ : Types.type_expr }
+  end
+
+  module Constructor : sig
+    type args = Tuple of Types.type_expr list | Record of Field.t list
+    type t = { name : string; args : args }
+  end
+
+  type param = { type_expr : Types.type_expr }
```
NathanReb, line 11: This indirection is not necessary atm! I think we'd better remove it as it simplifies the PR since normalize won't have so many modifications.

lib/diff.ml
```diff
@@ -1,39 +1,271 @@
 open Types
 open Stddiff
+open Intermed
```
NathanReb, line 3: Let's not stack up global opens. This one is only defining types, we can easily go around using the fully qualified names by using type annotations where it matters!

lib/convert.ml
```diff
@@ -0,0 +1,58 @@
+module T = Types
```
NathanReb, line 1: I think the whole `Convert` module could simply be moved to `Intermed`, having `a from_type_declaration` function in `Intermed.Type_decl`.

lib/diff.ml
```diff
@@ -1,39 +1,271 @@
...
+      @ [ reference ])
+      (List.map (fun param -> param.TypeDecl.type_expr) normed_cur @ [ current ])
+  then None
+  else Some { reference; current }
 
-and label = {
-  label_type : type_expr maybe_changed_atomic;
-  label_mutable : (Asttypes.mutable_flag, field_mutability) maybe_changed;
-}
+module TypeDecl = struct
```
NathanReb, line 18: Snake_case please!

lib/diff.ml
```diff
@@ -1,39 +1,271 @@
...
+      | Same _, None -> None
+      | _, None -> Some { mutable_; type_ = Same reference.type_ }
+      | _, Some type_change -> Some { mutable_; type_ = Changed type_change }
+  end
+
+  module Constructor = struct
+    type args =
+      | Record of (F.t, Field.t) map
+      | Tuple of type_expr maybe_changed_atomic_entry list
+      | Unshared of C.args atomic_modification
```
NathanReb, line 56: I think we've been going for `Atomic` for these, it would be best to use the same convention everywhere!

lib/text_diff.ml
```diff
@@ -54,40 +54,138 @@ let pp_ diff_printer =
...
+      Inline_hunks
+        (match ihunks with
+        | [] -> ihunks
+        | ihunk :: ihunks' ->
+            (match ihunk with
+            | Icommon s -> Icommon (indent amount s)
+            | Iconflict { iorig; inew } ->
+                Iconflict
+                  {
+                    iorig = Option.map (indent amount) iorig;
```
NathanReb, line 80: That would store the extra spaces in the conflict and we don't want that. You can simply append an `Icommon s` where `s` is the indentation.

Another solution would be to store the indentation level directly into the hunk so that this function does not have to allocate new strings and this can be dealt with when actually displaying the diff.

lib/text_diff.ml
```diff
@@ -54,40 +54,138 @@ let pp_ diff_printer =
...
+              params))
+
+let private_to_string private_ = if private_ then " private" else ""
+let mutable_to_string mutable_ = if mutable_ then " mutable" else ""
+
+let field_to_string field =
+  let open Intermed.TypeDecl.Field in
+  let mutable_string = mutable_to_string field.mutable_ in
+  let type_expr_string = type_expr_to_string field.type_ in
+  Printf.sprintf " %s%s : %s;" mutable_string field.name type_expr_string
```
NathanReb, line 118: There's someting a bit off there. If the field is mutable, then there will be two spaces at the start. I'm not sure this is intended.

The `to_string` functions should probably not handle surrounding spaces there as it otherwise becomes hard to track across the code. The caller should probably handle the spaces to separate the syntax elements.

tests/api-diff/type_kind_tests.t
```diff
@@ -65,9 +65,9 @@ Run the api-watcher on record and open type kinds cmi files
 
   $ api-diff ref_record_kind.cmi ref_open_kind.cmi
   diff module Ref_open_kind:
-   type t =
+  -type t =
+  +type t = ..
   -  { a : int; b : float; }
-  +  ..
```
NathanReb, line 71: This is incorrect and should probalby be shown as:
```
- type t =
-  { a : int ...
+ type t = ..
```
I think it would also be fine, and probably simpler to display it as:
```
 type t =
-  { a : int ...
+  ..
```

lib/text_diff.ml
```diff
@@ -269,80 +269,41 @@ and process_type_params_diff params_diff =
...
+      [ Icommon " ="; Iconflict { iorig = Some " private"; inew = None } ]
+
+and process_type_expr_diff diff =
+  let module S = Stddiff in
+  match diff with
+  | S.Same type_expr -> Icommon (" " ^ type_expr_to_string type_expr)
+  | S.Changed { S.reference; current } ->
+      Iconflict
+        {
+          iorig = Some (" " ^ type_expr_to_string reference);
```
NathanReb, line 290: The extra space can be made an `Icommon` hunk!

lib/intermed.ml
NathanReb: I agree but this won't happen overnight I think so we shouldn't bother with it until then.

lib/text_diff.ml
azzsal: > Another solution would be to store the indentation level directly into the hunk so that this function does not have to allocate new strings and this can be dealt with when actually displaying the diff.

I totally agree, but I think we should take of refining the textual diff types to take care of spacing and indentation in separate PR, and maybe defining our own pretty printers for our IR of type declarations.

NathanReb (commented):
Looks great! I have a few minor comments but other than those it should be good to go!

I think there a few conflicts so we should try to rebase this early and merge it next to avoid going through the rebasing again!

PR #134, opened 2025-03-05
Fix stack overflow

azzsal:


NathanReb:
Could you add a bit of documentation in typing_env explaining how the subst is built and why? This will be helpful for review and in the future if we have to modify this bit of the code !

lib/diff.ml
```diff
@@ -354,19 +380,22 @@ let class_type_item ~typing_env ~name
...
-let rec items ~reference ~current =
-  let env = Typing_env.for_diff ~reference ~current in
-  let ref_items = extract_items reference in
-  let curr_items = extract_items current in
+let rec items ~reference ~current ~typing_env =
+  let ref_items =
+    extract_items reference ~sub:false typing_env.Typing_env.subst
+  in
+  let curr_items =
+    extract_items current ~sub:true typing_env.Typing_env.subst
```
NathanReb, line 388: It'd be great to have a comment here explaining why this one needs to have `~sub:true`. It would also probably be great to have a doc comment on `extract_items` explaining what the function does and what the `~sub` argument controls.

azzsal: I updated the comment in the `lib/typing_env.mli`, I think that's enough to understand `sub` flag in `extract_items`.

lib/typing_env.ml
```diff
@@ -1,50 +1,121 @@
...
 
-let extract_non_alias_types signature =
+type t = { env : Env.t; subst : Subst.t }
+type subst_kind = Type | Module | Modtype [@@deriving ord]
+
+module Subst_item_map = Map.Make (struct
+  type t = subst_kind * string [@@deriving ord]
+end)
+
+let uniq_signature_ids ~reference ~current =
```
NathanReb, line 10: There a quite a lot of substitution going on so I think providing a bit of context for this function would help.

I am correct that this will traverse the whole `current` sig, registering a subst for any ID that already exist in the `reference` signature to a fresh ID and then running the substitution to return a new signature with IDs updated?

I think we could both have a better name for this function and document it to describe this behaviour in a few lines so we don't have to read the code to understand what it does.

How about `replace_matching_ids ~reference ~current` or something along those lines?

lib/typing_env.ml
```diff
@@ -1,50 +1,121 @@
...
-          (Btype.newgenty (Tconstr (ref_path, curr_decl.type_params, ref Mnil)));
-    }
-  else curr_decl
+    Subst_item_map.empty signature
 
-let set_type_equalities ~reference ~current =
-  let ref_non_alias_types = extract_non_alias_types reference in
-  List.map
-    (fun item ->
+let items_subst ~reference ~current =
```
NathanReb, line 90: This function is creating a `Subst.t` that's meant to be used to replace types and other such items that should be considered equal is that correct?

I think this could also use a more meaningful name and a tiny bit of documentation!

tests/api-diff/type_tests.t
```diff
@@ -68,6 +68,8 @@ Run api-watcher on the two cmi files, there should be a difference
 
   $ api-diff ref.cmi modify_type.cmi
   diff module Modify_type:
+  -val f : t -> string
```
NathanReb, line 71: Is that `t` the one we report a change for below?

Shouldn't it be expanded to `int` and `float` respectively? Or this not fixed by this PR?

If it's indeed not fixed we should add a mention of it in the tests saying that it exhibits a buggy behaviour!

NathanReb: This is also not related to this PR and therefore something we should fix separately but we should take care in not printing the value before the type as it is dependent on it! It would be worth opening an issue about that!

NathanReb (commented):
This looks great! Awesome work!

Most of my comments are about documenting or naming things to make it clear what are the different steps involved and why they are necessary.

I also spotted a couple quirks in the tests but besides that it looks good!

I'm also wondering, since we still use a single global env, do you think it might make sense for `Typing_env.for_diff` to return the current signature with the "type equalities" subst already applied or is there a specific reason we do it individually for items when assembling the item map in `Diff`?

lib/typing_env.ml
azzsal: Yes. It think it was called ```pair_components``` in the compiler libs, Is that a good name ?

tests/api-diff/type_tests.t
azzsal: Yes it should be expanded. But `Ctype.full_expand` doesn't appear to recurse over the type expr, so we have to recurse ourselves, I think it would be better to handle that when we start diffing the core type.

lib/typing_env.ml
azzsal: > I am correct that this will traverse the whole current sig, registering a subst for any ID that already exist in the reference signature to a fresh ID and then running the substitution to return a new signature with IDs updated?

Exactly!

tests/api-diff/type_tests.t
NathanReb: Agreed!

NathanReb: Could you still add a mention the test is not exhibiting the expected behaviour please?

lib/typing_env.ml
NathanReb: Yeah that sounds good!

tests/api-diff/value_tests.t
```diff
@@ -69,3 +69,27 @@ Run api-diff and check the output
   +val f : t -> t
   
   [1]
+
+### Value referencing an abstract type
+
+api-diff should be able to tell two, alias types from both versions of the
```
NathanReb, line 75: Don't you mean the opposite? We need to make sure non alias types are treated as equal.

Alias types are just resolved so here we're essentially comparing `int` to `int`. I think a more meaningful test would be something like
```
type t = {a: int}
val x : t
```
VS
```
type t = {a: float}
val x : t
```
Wouldn't it? There we would expect a change reported for the type but not for the value and for this to work we have to tell the compiler somehow that `t1` (from ref) is the same as `t2` (from cur) which is done by substituting `t2` by `t1` using `Subst`.

Isn't that what we mean to test here? Alias types are not substituted as far as I can tell.

NathanReb (approved):
Awesome work! Let's merge!

PR #135, opened 2025-03-07
Initialize typing env with stdlib

azzsal:


azzsal:
There is one [failing test](https://github.com/ocaml-semver/ocaml-api-watch/blob/main/tests/api-watch/test_diff_module.ml#L92) related to functors, but we cannot detect that since there are supported and unsupported changes, but we can only detect one, not both. I will update the description of the test there.

tests/api-watch/test_diff_module.ml
```diff
@@ -118,8 +119,7 @@ let%expect_test "Submodules with different functor types" =
   Format.printf "%a" pp_diff_option result;
   [%expect
     {|
-    Some (Module Main: {Modified (Supported [ Module F: {Modified (Unsupported)};
```
NathanReb, line 121: Why did this change? There is no apparent reasons why we couldn't report this change anymore!

lib/diff.ml
```diff
@@ -429,11 +429,13 @@ and signatures ~reference ~current =
     items ~reference:modified_reference ~current:modified_current ~typing_env
   with
   | [] -> (
+      Compmisc.init_path ();
+      let initialized_env = Compmisc.initial_env () in
```
NathanReb, line 433: Might be worth extracting a function to build a fresh environment with the stdlib (and later, any external dependencies) loaded.

I think `Typing_env` would be a good place to write this function, especially since it will use it internally as well!

tests/api-watch/test_diff_module.ml
azzsal: Sorry my bad, we do support functors, but if there was a change was just report an unsupported change on them. It seems that `Includemod.modtypes` doesn't detect a change on functor parameters anymore for some reason. I need to investigate this further.

lib/typing_env.ml
```diff
@@ -117,12 +125,15 @@ let pair_items ~reference ~current =
       | _ -> subst)
     Subst.identity current
 
-let for_diff ~reference ~current =
+let initialized_env () =
   Compmisc.init_path ();
```
NathanReb, line 129: We should be a bit careful with this, I think it should only be called once. Might be harmless to call it several times but to err on the safe side I'd make sure we only do it once!

lib/typing_env.ml
```diff
@@ -45,19 +45,27 @@ let replace_matching_ids ~reference ~current =
...
-        | Sig_modtype (id, mtd, v) as sig_modtyp_decl -> (
+        | Sig_modtype (id, mtd, v) -> (
             match Env.find_modtype_index id ref_env with
             | Some _ ->
                 let new_id = Ident.rename id in
                 ( Subst.add_modtype id (Mty_ident (Pident new_id)) subst,
                   Sig_modtype (new_id, mtd, v) :: lst )
-            | None -> (subst, sig_modtyp_decl :: lst))
+            | None ->
+              (** This is a special case for functor paramters, should be
```
NathanReb: Could you provide a bit more explanation here? I think you found a pretty nasty bug and it'd be great to have it documented in the history somewhere at least!

We might even add that to the documentation of the whole function as a justification to why rewriting IDs is necessary as long as we rely on the compiler's functions.

lib/typing_env.ml
azzsal: I did it with closures, I was wondering if it would be any better to do it with lazy values ?
`let initialized_env = lazy (Compmisc.init_path (); Compmisc.initial_env ())`

lib/typing_env.ml
azzsal: I added some documentation, feel free to change it if it's not clear enough.

lib/typing_env.ml
NathanReb: Both are fine by me!

NathanReb (approved):
Perfect, thanks!

PR #136, opened 2025-03-09
Fix word-diff command line options

azzsal:


bin/api_diff.ml
```diff
@@ -4,16 +4,12 @@ type mode = Unwrapped | Wrapped of string | Cmi
...
-      Error
-        "Invalid argument. --word-diff takes an optional mode arguemnt.\n\
-         The mode can be either color or plain. If no mode is provided, it \
-         defaults to plain."
+let display_mode word_diff color =
+  match (word_diff, color) with
+  | false, false -> Ok Line
+  | true, false -> Ok (Word Plain)
+  | false, true -> Ok (Word Color)
+  | true, true -> Error "--word-diff and --color cannot be specified together"
```
NathanReb, line 12: Why not?

Also without `--word-diff` but with `--color` you get the `Word Color` output? I don't think that's right or that it should work like this.

I think it'd be best to have the following:
1. no flags: line based, no colors
2. `--colors`: line based, colors
3. `--word-diff`: word based, no colors (i.e. `Word Plain`)
4. `--word-diff --colors`: word based, colors (i.e. `Word Color`)

There are two things that bother me here:
1. Colored output is not the default anymore
2. You can't have both colors and plain word diff which is something git supports

I'd suggest we instead swap the `--color` flag to a `--no-colors` and document it as follows:
```
Force the diff output to contain change markers visible in terminals that don't support coloring, e.g. showing word diffs as [-removed-]{+added+} instead of simply highlighting them in red and green.
```

We could then still always output everything colored and let `Fmt` detect whether it can or cannot output colors.

That would turn our pattern matching into:
1. no flags: line based, colored if possible
2. `--no-colors`: line based, colored if possible (same as above, line based diff already contains markers visible without colors)
3. `--word-diff`: word based, `Word Color` style, colored if possible
4. `--word-diff --no-colors`: word based, `Word Plain` style, colored if possible

What do you think?

NathanReb (commented):
I made a suggestion regarding the flags and how they should interact with each other. Let me know what you think!

This should have a changelog entry as it introduce user visible changes: the new output mode and related flags.

bin/api_diff.ml
azzsal: I find that `--no-colors` a bit confusing. An end user would expect it would output the diff without colors, but it would be colored if the terminal supports colors (most likely). 

What about something like this:

1. ```--line=color``` : line based, with color highlighting.
2. ```--line=plain``` : line based, with [-removed-] and {+added+} markers (we could document that this should be used if the terminal does not support colors), colored if possible.
3. ```--word=color``` : word based, with colors.
4. ```--word=plain``` : word based, with [-removed-] and {+added+} markers, colored if possible.

NathanReb: Yes I agree the name isn't the best but I think the documentation is pretty clear that it does not control whether we color the output isn't it?

Maybe we can find a better name that encapsulates this idea, something like `--inline-markers`?

Maybe the solution is to keep the current `--word-diff` option.

I guess I'd still be curious to see if we could have something like:
```
$ api-diff --inline-markers ref.cmi curr.cmi
- type t = [-int-] list
+ type t = {+float+} list
```

So that line diff still has inline highlighting even without colors.

What do you think?

NathanReb: Sorry, to clarify:
1. What do you think of a different flag name such as `--inline-markers`? The documentation would state
```
Add text markers to the output to highlight inline changes. Deleted parts are wrapped between [- and -] and added parts between {+ and +}. Useful for terminals or outputs that don't support colors.
```
2. What do you think about making this work with the line based diff as well?

azzsal: So with this we would have four options:

1. without any flags: line based with color highlighting
2. `--inline-markers`: line based with markers (colored if possible)
3. `--word-diff ``: word based with color highlighting
4. `--word-diff --inline-markers`: word based with markers (colored if possible)

It look good to me, though I don't like the fact we have to specify two separate flags to have such simple behavior.

NathanReb: There are two orthogonal properties (line VS word) and (pure color VS color and text based highlighting) so this is bound to require two flags or two options.

I admit I like flags better than options, especially when options only accept two distinct values. That means that with the option you need to remember both the option name and the possible values.

CHANGES.md
```diff
@@ -28,6 +28,7 @@
 - Remove duplicate items in class and class types (#105, @azzsal)
 - Fixed loading of modules whose signature is given by a path to a module type:
   `module X : Y` (#128, @panglesd)
+- Fixed command line flags associated with displaying the diff (#136, @azzsal)
```
NathanReb, line 31: This would have to go under Changed as we're not fixing a bug here, but modifying our interface and how we use existing CLI flags and options.

Since the original change has not been released yet I would merge this with the following entry:
```markdown
- Add word-level display of textual diffs (#131, @azzsal)
```
turning it into:
```markdown
- Add word-level display of textual diffs in `api-diff`, enabled with the `--word-diff` flag  (#131, #136, @azzsal)
- Add `--plain` flag to `api-diff` to use text markers for inline highlighting. Can be used when the output does not support colors (#136, @azzsal)
```

A changelog is meant to be read by users not only developers of the project, we need to give context and actionable info there.

bin/api_diff.ml
```diff
@@ -99,18 +104,16 @@ let run (`Word_diff word_diff) (`Color color) (`Main_module main_module)
 
 let named f = Cmdliner.Term.(app (const f))
 
-let color =
+let plain =
   let doc =
-    "Dispaly API changes in an inline word diff format, the inline changes are \
-     showed in colors."
+    "Add test markers to the output to highlight inline changes. Deleted parts \
```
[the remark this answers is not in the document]
NathanReb: Or maybe you meant `text`?

PR #139, opened 2025-03-13
Add fine grained diff of tuple type exprs

### LINKED ISSUE #138: Fine grained diff of tuple types
We'd like to be able to compute the structural diff between two tuple types, that is, between something like:
```ocaml
int * int
```
and
```ocaml
int * float
```
So that we know that the first component of the tuple is identical and the second contains a diff.

This should be able to deal with tuples recursively, meaning that it should be able to tell the exact diff between:
```ocaml
int * (bool * bool)
```
and
```ocaml
int * (bool * int)
```
i.e. that the diff is something like:
```ocaml
[Same int; Diff [Same bool; Diff {reference = bool; current = int}]]
```

Whenever we encounter a type expression we can't get a structural diff for, we default back to an atomic diff.

azzsal:
This should resolve #138.

### DIFF
--- a/lib/diff.mli
+++ b/lib/diff.mli
@@ -1,26 +1,29 @@
-type typ_exp =
+type type_expr =
   | Tuple of tuple
   | Atomic of Types.type_expr Stddiff.atomic_modification
 
-and tuple = (Types.type_expr, typ_exp) Stddiff.List_.t
+and tuple = (Types.type_expr, type_expr) Stddiff.List.t
 
 type type_modification = {
   type_kind : (Types.type_decl_kind, type_kind) Stddiff.maybe_changed;
   type_privacy : (Asttypes.private_flag, type_privacy) Stddiff.maybe_changed;
-  type_manifest : (Types.type_expr, typ_exp) Stddiff.Option_.t;
+  type_manifest :
+    ( Types.type_expr option,
+      (Types.type_expr, type_expr) Stddiff.Option.t )
+    Stddiff.maybe_changed;
   type_params :
     ( Types.type_expr list,
-      (Types.type_expr, typ_exp) Stddiff.List_.t )
+      (Types.type_expr, type_expr) Stddiff.List.t )
     Stddiff.maybe_changed;
 }
 
 and type_kind =
-  | Record_tk of (Types.label_declaration, label) Stddiff.Map_.t
-  | Variant_tk of (Types.constructor_declaration, cstr_args) Stddiff.Map_.t
+  | Record_tk of (Types.label_declaration, label) Stddiff.Map.t
+  | Variant_tk of (Types.constructor_declaration, cstr_args) Stddiff.Map.t
   | Atomic_tk of Types.type_decl_kind Stddiff.atomic_modification
 
 and label = {
-  label_type : (Types.type_expr, typ_exp) Stddiff.maybe_changed;
+  label_type : (Types.type_expr, type_expr) Stddiff.maybe_changed;
   label_mutable :
     (Asttypes.mutable_flag, field_mutability) Stddiff.maybe_changed;
 }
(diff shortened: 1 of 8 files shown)


lib/diff.mli
```diff
@@ -1,34 +1,38 @@
+type typ_exp =
```
NathanReb, line 1: I think we should adopt the convention that `Diff.x` represent the diff between two values of type `Types.x`. I'd rename this to `type_expr`.

lib/stddiff.ml
```diff
@@ -11,65 +11,71 @@ type 'a atomic_modification = { reference : 'a; current : 'a }
...
-  changed_map : ('item, 'diff) entry String_map.t;
-}
-
-type ('same, 'change) maybe_changed = Same of 'same | Changed of 'change
 type 'item atomic_entry = ('item, 'item atomic_modification) entry
+type ('same, 'change) maybe_changed = Same of 'same | Changed of 'change
 
-type 'same maybe_changed_atomic =
-  ('same, 'same atomic_modification) maybe_changed
+module List_ = struct
```
NathanReb, line 17: I think we can remove the trailing `_` here. `Stddif` shouldn't be globally open so it won't shadow stdlib's modules.

lib/stddiff.ml
```diff
@@ -11,65 +11,71 @@ type 'a atomic_modification = { reference : 'a; current : 'a }
...
+    in
+    let list_diff, all_same = aux reference current ([], true) in
+    if all_same then Same reference else Changed list_diff
+end
 
-type ('same, 'diff) option_ = ('same option, ('same, 'diff) entry) maybe_changed
-type 'same atomic_option = ('same, 'same atomic_modification) option_
-type ('same, 'diff) list_ = ('same list, 'diff list) maybe_changed
+module Option_ = struct
+  type ('a, 'diff) t = ('a option, ('a, 'diff) entry) maybe_changed
```
NathanReb, line 38: This one is wrapped in a `maybe_changed` but the `List` one isn't. We should probably stick to the same convention here.

I don't think `M.t` should wrap into a maybe changed. The `diff` function could return an `(option, t) maybe_changed` but the type itself is the diff type, i.e. the type that represents the change.

lib/text_diff.ml
```diff
@@ -482,18 +466,25 @@ and process_tuple_type_diff tuple_diff =
...
+  | Diff.Atomic { reference; current } ->
+      [
+        Iconflict
+          {
+            iorig = Some (type_expr_to_string reference);
+            inew = Some (type_expr_to_string current);
+          };
+      ]
+  | Tuple tuple_diff ->
+      (Icommon "(" :: process_tuple_type_diff tuple_diff) @ [ Icommon ")" ]
```
NathanReb, line 486: I don't think that's the right place for adding brackets or at least not unconditionally, the caller should probably be responsible for this.

This could be controlled by a labelled arg for instance!

tests/api-diff/tuple_tests.t
```diff
@@ -0,0 +1,48 @@
...
+We generate the .cmi file
+
+  $ ocamlc change.mli
+
+Run the api-watcher on the two cmi files, a diff should be reported between the first
+components of the two tuples
+
+  $ api-diff --plain ref.cmi change.cmi
+  diff module Change:
+  -type t = ([-int-] * int * int)
```
NathanReb, line 26: That's a good example showing why we shouldn't add brackets unconditionally!

NathanReb (commented):
This looks great!

I have a few comments regarding consistency of types and names but the overall refactoring looks good and the new types in `Stddiff` are a great addition.

The printing is a bit inconsistent but it shouldn't be too hard to fix!

We should probably add a test for tuples within tuples! That will be usefull both to test the brackets and that the diffing works correctly!

NathanReb (approved):
Looks great, thanks!

Let's merge!

PR #143, opened 2025-03-15
Add fine grained diff of arrow type exprs

### LINKED ISSUE #140: Fine grained diff of arrow types
Similarly to tuples, we should support diffing individual type expressions in an arrow type.

This should be done recursively so that we ultimately can identify changes in any of a function arguments or return type.

Anything that's not an arrow (or another part of `type_desc` we already diff) should be diffed using the fallback method for type expressions.

azzsal:
This should resolve #140.


lib/diff.ml
```diff
@@ -168,6 +197,73 @@ and type_exprs ~typing_env ~ref_params ~cur_params ~reference ~current =
...
     ~reference ~current
 
+and arrow ~typing_env ~ref_params ~cur_params ~reference ~current =
+  let arg_type lbl typ =
+    match lbl with
+    | Asttypes.Nolabel | Labelled _ -> typ
+    | Optional _ -> (
+        match Types.get_desc typ with
+        | Tconstr (_, [ te ], _) -> te
+        | _ -> assert false)
```
NathanReb, line 207: This function should be given a proper name and a bit of documentation as to why it's needed!

What do you think of `unwrap_optional_arg`?

lib/text_diff.ml
```diff
@@ -475,6 +475,71 @@ and process_tuple_type_diff diff =
...
+    | Changed type_expr -> process_type_expr_diff type_expr
+  in
+  let return_type_ihunks =
+    match arrow_diff.return_type with
+    | Same type_expr -> [ Icommon (type_expr_to_string type_expr) ]
+    | Changed type_expr -> process_type_expr_diff type_expr
+  in
+  List.concat
+    [
+      [ Icommon "(" ];
```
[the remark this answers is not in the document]
NathanReb: We probably need to change the `paren:bool` argument to some sort of
``context:[`Tuple | `Larrow | `Rarrow]``

tests/api-diff/arrow_tests.t
```diff
@@ -0,0 +1,47 @@
+Here we generate a `.mli` file with an arrow type:
```
NathanReb, line 1: It might be best to merge the type expression diffing tests together under `fine_grained_type_expr.t` so we have a natural place to test how they all tie together, e.g. whether a tuple within an arrow is well displayed or the otherway around.

NathanReb (commented):
This looks good!

Properly handling the conversion to textual diff is going to require a bit of work but getting it right is important as showing diffs between type expressions is going to be at the core of what we do.

I think it could use a bit more test cases as well, e.g. checking diffs in the result type, or in a second argument are properly handled. Seeing how tuples within arrows and the other way around are handled is important as well as it will help us not to break either the diffing or the text diff conversion in the future.

lib/text_diff.ml
```diff
@@ -447,100 +444,109 @@ and process_cstr_diff name cstr_diff =
...
+  let arrow_hunks =
+    List.concat
+      [
+        arg_label_ihunks;
+        arg_type_ihunks;
+        [ Icommon " -> " ];
+        return_type_ihunks;
+      ]
+  in
+  if context = `Tuple || context = `Larrow then
```
NathanReb, line 545: I think I'd prefer a pattern matching here:
```suggestion
  match context with
  | `Tuple | `Larrow ->
```

lib/text_diff.ml
```diff
@@ -447,100 +444,109 @@ and process_cstr_diff name cstr_diff =
...
+    (Icommon "(" :: arrow_hunks) @ [ Icommon ")" ]
```
NathanReb, line 546: These shows up in a few places and I think we'll need again so me might want to extract a function for this, something like `wrap_with_bracket_hunks`.

lib/text_diff.ml
NathanReb: It's also best to use a pattern-matching here because, by making it exhaustive, we'll get a compiler warning if we ever add new possible contexts!

NathanReb (commented):
Looks great! I just have a couple of very minor comments but once that's fixed we can merge!

lib/text_diff.ml
```diff
@@ -542,9 +541,11 @@ and process_arrow_type_diff ~context arrow_diff =
         return_type_ihunks;
       ]
   in
-  if context = `Tuple || context = `Larrow then
-    (Icommon "(" :: arrow_hunks) @ [ Icommon ")" ]
-  else arrow_hunks
+  match context with
+  | `Tuple | `Larrow -> parenthesize arrow_hunks
+  | _ -> arrow_hunks
```
NathanReb, line 546: This case should be explicit, if we use a wildcard, we won't get warned by the compiler if we pass a new context down the chain and might forget to handle it correctly here! Same for the one above!

PR #146, opened 2025-03-17
Add support for extension constructors

### LINKED ISSUE #145: Detect addition, removal and modifications of extensible types' variants

azzsal:
This should resolve #145.


NathanReb (commented):
This looks good! I have a few minor suggestions but other than that it looks good.

I think adding test for handling of extensible variant types parameters, regardless of how we choose to handle them, would be good here!

NathanReb (approved):
Looks great, thanks! There's just a changelog conflict to fix but then it's good to go!

PR #148, opened 2025-03-21
Add fine grained diffing of type constructors

### LINKED ISSUE #147: Fine grained diff of type constructors
This is the `Tconstr` variant in from `type_desc`.

This has two parts:
- The list of type parameters
- The type itself, represented as a `Path.t`

To properly diff this we need to be able to resolve paths as the path not matching does not necessarily mean the types are different.

A first approach we suggest here is to always fully resolve type expressions. E.g. if we consider the following interface:
```ocaml
type 'a r = ('a, string) result

type s = int r

type t = s
```
`t` should be resolved to `(int, string) result` before being diffed. Alternatively, if a type resolves to a concrete type declaration (e.g. a record or variant type) it should be resolved to the type declaration ID.

A good first approach is to fully resolve both `Tconstr` whenever we are diffing them and to in fact diff the fully resolved types.

This will always be correct, even if the resulting diff might look a bit odd as it won't respect the aliases used across the documentation for instance. This can be dealt with later by trying to minimize the resolution to the first matching types between the two `Tconstr`.

azzsal:
This should resolve #147.


tests/api-diff/module_tests.t
```diff
@@ -107,14 +107,14 @@ Run api-diff and check the output
   diff module Modified_module:
   -val a : string -> int
   +val a : string -> float
-  -val f : int -> string
+  -val f : int ->  string
```
NathanReb, line 110: There's an unwanted extra space here

tests/test_helpers/test_helpers.mli
```diff
@@ -3,3 +3,6 @@ val compile_interface : string -> Types.signature
 
 val first_type_declaration :
   Types.signature -> (Ident.t * Types.type_declaration) option
+
+val get_path : Types.type_expr -> Path.t
+val get_args : Types.type_expr -> Types.type_expr list
```
NathanReb, line 8: Since we always need both together, how about:
```suggestion
val get_tconstr : Types.type_expr -> Path.t * Types.type_expr list
```

NathanReb (commented):
This looks to be working well, good job!

There are a few things that needs to improved code-wise but it looks promising!

tests/api-diff/modified_variant_type_tests.t
```diff
@@ -195,10 +195,10 @@ Run the api-watcher on the two cmi files
   $ api-diff shapes2.cmi add_component.cmi
   diff module Add_component:
    type shape =
-  -  | Circle of point * int
-  +  | Circle of point * int * int
-  -  | Rectangle of point * point
-  +  | Rectangle of point * point * int
+  -  | Circle of float * float * int
+  +  | Circle of float * float * int * int
```
NathanReb, line 199: we're missing brackets here and this makes the overall expression incorrect! Also, since `point` is unchanged, I'd expect our code to produce `Same point` rather than `Same foat * float` right?

tests/api-watch/test_subst_params.ml
```diff
@@ -0,0 +1,70 @@
...
+  let expanded_type_expr =
+    Typing_env.expand_tconstr ~typing_env:env ~path ~args
+  in
+  match expanded_type_expr with
+  | Some _ -> assert false
+  | None ->
+      Printtyp.type_expr Format.std_formatter v2_type;
+      [%expect {| (string, int) t2 |}]
+
+let%expect_test "test_expand_tconstr_on_type_not_in_the_env" =
```
NathanReb, line 60: If this is testing type outside the env, I think I'd use something else than `list` that we usually expect to be in the env as it's provided by the standard library. Using something like `not_in_env` as a type name there would make it entirely clear!

NathanReb (commented):
This looks much better! Thanks!

There still a few quirks remaining but then this should be good to go!

I'd a test for the expansion of `('a, 'b) t` when `t` is defined as `type t = 'a * 'b` as well to test a bit how it behaves when we expand to something that's not a Tconstr.

If we're going for the full expand function as well, it'd be good to add unit tests for this function as well!

tests/api-diff/module_tests.t
azzsal: I will take care of this in #152.

lib/diff.ml
```diff
@@ -191,6 +218,31 @@ let rec type_expr ~typing_env ?(ref_params = []) ?(cur_params = []) reference
...
+          type_exprs ~typing_env ~ref_params ~cur_params ~reference:ref_args
+            ~current:cur_args
+        in
+        match type_exprs with
+        | Same same_params -> Same same_params
+        | Changed change -> Changed (Modified change))
+  in
+  match (path, args) with
+  | Same _, Same _ -> Same reference
+  | _ -> Changed { path; args }
```
NathanReb, line 244: Why not returning `Changed (Constr {path; args})` directly here?

lib/typing_env.ml
```diff
@@ -148,23 +148,27 @@ let for_diff ~reference ~current =
...
+
+let rec fully_expand_type_expr ~typing_env ~type_expr ~path ~args =
+  let exp_res = expand_tconstr ~typing_env ~path ~args in
+  match exp_res with
+  | None -> type_expr
+  | Some expr -> (
+      match Types.get_desc expr with
+      | Tconstr (path, args, _) ->
+          fully_expand_type_expr ~typing_env ~type_expr:expr ~path ~args
+      | _ -> expr)
```
NathanReb, line 171: `type_expr` is an internal here used to make the function tail recursive. I'd suggest the following:
```suggestion
let fully_expand_tconstr ~typing_env ~path ~args =
  let rec aux last path args =
    match expand_tconstr ~typing_env ~path ~args with
    | None -> last
    | Some expr ->
      (match Types.get_desc expr with
      | Tconstr (path, args, _) ->
        aux (Some expr) path args
      | _ -> Some expr)
    in
    aux None path args
```
We then deal with the `None` case at the call site with some `Option.get_value ~default:` or something along those lines.

lib/diff.ml
```diff
@@ -194,13 +188,31 @@ let rec type_expr ~typing_env ?(ref_params = []) ?(cur_params = []) reference
...
+          | Stddiff.Same _ -> Stddiff.Same current
+          | Changed change -> Changed change)
+      | _, _ -> (
+          let diff =
+            type_expr ~typing_env ~ref_params ~cur_params expanded_ref
+              expanded_cur
+          in
+          match diff with
+          | Same _ -> Same current
+          | Changed change -> Changed change))
```
NathanReb, line 215: Let's move this into an `expand_and_diff_tconstr` to keep the type_expr function a bit more readable and reasonably sized.

lib/typing_env.mli
```diff
@@ -22,8 +22,25 @@ val for_diff :
...
 val fully_expand_type_expr :
-  typing_env:t -> type_expr:Types.type_expr -> Types.type_expr
+  typing_env:t ->
+  type_expr:Types.type_expr ->
+  path:Path.t ->
+  args:Types.type_expr list ->
+  Types.type_expr
 (** Expand the given [type_expr], looking up the environment for an existing
     alias and applying the type parameters as needed at each step.
 *)
```
NathanReb, line 46:
```suggestion
(** Recursively expand the given path and args, looking up the environment for an existing
    alias and applying the type parameters as needed at each step, until expanded to anything but a [Tconstr ...].
*)
```

NathanReb (commented):
Could you add at least one test for the full expansion function?

tests/api-watch/test_fully_expand_type_expr.ml
```diff
@@ -39,28 +39,34 @@ let%expect_test "test_fully_expand_type_expr_on_alias_types" =
...
+      Printtyp.type_expr Format.std_formatter expr;
+      [%expect {| (string * int) list |}]
 
 let%expect_test "test_fully_expand_type_expr_on_nominal_types" =
   let v2_type = String_map.find "v2" value_map in
   let path, args = get_tconstr v2_type in
   let expanded_type_expr =
-    Typing_env.fully_expand_type_expr ~typing_env:env ~type_expr:v2_type ~path
-      ~args
+    Typing_env.fully_expand_tconstr ~typing_env:env ~path ~args
```
NathanReb, line 54: Sorry, that's not what I meant. We should keep the existing single step expansion tests and add an extra one that tests multiple levels of expansion.

azzsal: I already test both. [Here](https://github.com/ocaml-semver/ocaml-api-watch/pull/148/files#diff-9b8adf3218f9ac231d6ff38ed29f2aa45cf21b5438ef7873023dd0370eb7680c) I test the single step expansion, and [here](https://github.com/ocaml-semver/ocaml-api-watch/pull/148/files#diff-2045d9faff87dc0cbd188de3f528977c93fe773efe8a19cfe7ede9d668ad7f97) I test the full expansion.
Could you provide an example with a case that you want to test ?

NathanReb: my bad, the diff was shown as just a rename!

NathanReb (approved):

PR #153, opened 2025-04-01
Fix expansion of type constructors

azzsal:


lib/diff.ml
```diff
@@ -230,7 +241,14 @@ and expand_and_diff_tconstr ~typing_env ~ref_params ~cur_params ~reference
       in
       match constr with
       | Stddiff.Same _ -> Stddiff.Same cur_expr
-      | Changed change -> Changed change)
+      | Changed change -> (
+          match
+            type_expr ~typing_env ~expand:false ~ref_params ~cur_params
```
NathanReb, line 246: You re-introduced an extra parameter used for recursion here. I'm not sure we really need the `expand` parameter. As we were doing before we can run the expansion mechanism and match over the resulting expressions:
- if both are `Tconstr`, we diff them with `constr` directly, without going through `type_expr` again
- if they aren't, we forward them to `type_expr`

lib/typing_env.mli
```diff
@@ -35,11 +35,12 @@ val expand_tconstr :
...
 
-val fully_expand_tconstr :
+val expansion_lst :
   typing_env:t ->
+  type_expr:Types.type_expr ->
   path:Path.t ->
   args:Types.type_expr list ->
-  Types.type_expr option
+  (Types.type_expr * Types.type_declaration option) list
 (** Recursively expand the given path and args, looking up the environment for
```
NathanReb, line 44: The doc comment is not correct anymore!

lib/typing_env.ml
```diff
@@ -158,18 +158,21 @@ let expand_tconstr ~typing_env ~path ~args =
...
+    | None -> acc
+    | Some (expr, td) -> (
         match Types.get_desc expr with
-        | Tconstr (path, args, _) -> aux (Some expr) path args
-        | _ -> Some expr)
+        | Tconstr (path, args, _) -> aux ((expr, Some td) :: acc) path args
+        | _ -> (expr, Some td) :: acc)
   in
-  aux None path args
+  aux [ (type_expr, None) ] path args
```
NathanReb, line 175: This is a bit confusing, what does the return type of this function represent? What are the left and right hand side of the pairs?

lib/diff.ml
```diff
@@ -239,6 +257,29 @@ and expand_and_diff_tconstr ~typing_env ~ref_params ~cur_params ~reference
...
+  let ref_hd = List.hd reference in
+  let cur_hd = List.hd current in
+  aux (fst ref_hd, fst cur_hd) reference current
+
+and def_match ~typing_env ~reference ~current =
+  match (reference, current) with
+  | None, None -> false
+  | Some ref_td, Some cur_td
+    when Option.is_none
+           (type_declarations ~typing_env ~name:"" ~reference:ref_td
```
NathanReb, line 278: We're diffing the whole type declaration here? Isn't that overkill considering we only want to compare params and manifest? If there is no manifest there is no point comparing the rest of the definition as we should treat them as equal based on their path, e.g. if a record `M.N.t` from both versions of the API changed, we still want to treat it as equal. I don't think the code would be correct in this case.

Here we should think of an example to test this.
Reference:
```ocaml
type ('a, 'b) t = { a : 'a; b : 'b }
type 'a t1 = ('a, string) t
type t2 = int t1
val x : t2
```
Current:
```ocaml
type ('a, 'b) t = { a : 'a; b : 'b; flag : bool }
type 'a t1 = ('a, string) t
type t2 = float t1
val x : t2
```
In this example, I think the diff for the type of `x` should be shown as `[-int-]{+float+} t1`.

tests/api-watch/test_expansion_lst.ml
```diff
@@ -0,0 +1,97 @@
+open Api_watch
```
NathanReb, line 1: For unit tests we don't need one file per function tested. One unit test file per module (as in .ml file) is enough. Can you merge the two together?

lib/typing_env.mli
```diff
@@ -35,11 +35,12 @@ val expand_tconstr :
...
 *)
 
-val fully_expand_tconstr :
+val expansion_lst :
   typing_env:t ->
+  type_expr:Types.type_expr ->
   path:Path.t ->
   args:Types.type_expr list ->
-  Types.type_expr option
+  (Types.type_expr * Types.type_declaration option) list
```
NathanReb, line 43: This needs to be clarified a bit, I'm not sure how to interpret the option and I don't think we need the whole type declaration either.

lib/typing_env.ml
```diff
@@ -158,18 +158,21 @@ let expand_tconstr ~typing_env ~path ~args =
...
       | Some type_expr ->
-          Some (Ctype.apply typing_env td.Types.type_params type_expr args))
+          let expanded_expr =
+            Ctype.apply typing_env td.Types.type_params type_expr args
+          in
+          Some (expanded_expr, td))
 
-let fully_expand_tconstr ~typing_env ~path ~args =
-  let rec aux last path args =
+let expansion_lst ~typing_env ~type_expr ~path ~args =
```
NathanReb, line 166: It'd be good if we could avoid passing the type_expr here, I think it ends up messing with the return type more than anything an clutters a bit the function when we don't really need it.

We can see it as returning the expansion chain from the tconstr made of `path` and `args`. The caller could eventually use this and the type_expr it used to get the path and args to build another list, more suited to their own need. In other words, I think injecting the type_expr here is the caller code leaking into the implementation of this function.

NathanReb (commented):
This is looking good! Indeed it will likely need a bit more tests, at least for the cases we expect to find commonly in the wild.

There are a few things that can be simplified but it seems to be working well, congrats!

Issue #84, opened 2024-10-01
Disambiguate types


reynir:
I am writing this issue based off of the Outreachy demo earlier this year. Please correct me if the following is not the case anymore.

Inside a nested module `Foo.Bar.Baz` there might be a value `doit` with the following type:

```OCaml
val doit : t -> unit
```

In a diff without context it's unclear if `t` refers to `Foo.Bar.Baz.t`, `Foo.Bar.t` or `Foo.t`. I think in error messages I have seen `t/1` and `t/2` to precise what `t` is referred to. Alternatively, the type could be fully qualified, e.g. `val doit : Foo.Bar.t -> unit`.

NathanReb:
Yes you're right that this is important and definitely something we'll be aiming for in the output of the diff.

There is still quite a bit of work to do to properly set the typing environment for our diffing needs as we went with the bare minimum for the first prototype. For example it will fail if your API refers to types defined by another library, external to the one you're diffing.

Once we properly set the typing env, always using the fully qualified name seems like a good idea. For the sake of readibality we could also consider reserving non qualified names for types that are defined in the module itself.

PR #155, opened 2025-04-11
Fix env setup to handle cross module references

### LINKED ISSUE #154: Fix environment setup
Currently, each corresponding pair of signatures are diffed in their own separate environment. This creates problems if one signature references an item outside of its scope. We need to have the environment setup only once before diffing signatures, and diff the signatures in this shared environment.

**Minimal reproduction with .cmi files:**

```
$ cat > ref.mli << EOF
> module M : sig
>  type t = int
> end
> module N : sig
>  val x : M.t
> end
> EOF

$ ocamlc ref.mli

$ cat > cur.mli << EOF
> module M : sig
>  type t = float
> end
> module N : sig
>  val x : M.t
> end
> EOF

$ ocamlc cur.mli

$ dune exec -- api-diff ref.cmi cur.cmi
```
No diff will be reported on `val x : M.t`

**Minimal reproduction with wrapped libraries:**

```
$ mkdir ref_lib && cd ref_lib
$ cat > dune-project << EOF
> (lang dune 2.9)
> EOF

$ cat > dune << EOF
> (library
>  (name library))
> EOF

$ cat > m.ml << EOF
> type t = int
> EOF

$ cat > n.ml << EOF
> type u = M.t
> EOF

$ dune build

$ cd .. && mkdir cur_lib && cd cur_lib
$ cat > dune-project << EOF
> (lang dune 2.9)
> EOF

$ cat > dune << EOF
> (library
>  (name library))
> EOF

$ cat > m.ml << EOF
> type t = float
> EOF

$ cat > n.ml << EOF
> type u = M.t
> EOF

$ dune build

$ cd .. && dune exec -- api-diff --main-module=library ../ref_lib/_build/default/.library.objs/byte ../cur_lib/_build/default/.library.objs/byte
```
Same as with .cmi files, there isn't a diff reported on `type u = M.t`, while there should be one.

azzsal:
This should resolve #154

### DIFF
--- a/lib/diff.ml
+++ b/lib/diff.ml
@@ -749,7 +749,7 @@ and signatures ~typing_env ~reference ~current =
 
 let interface ~module_name ~reference ~current =
   let modified_reference, modified_current, typing_env =
-    Typing_env.for_diff ~reference ~current
+    Typing_env.for_diff ~module_name ~reference ~current
   in
   (*Typing_env.pp Format.std_formatter typing_env;*)
   let sig_out =
--- a/lib/diff.ml
+++ b/lib/diff.ml
@@ -711,23 +711,29 @@ and module_type ~typing_env ~name ~ref_module_type ~current_module_type
 
 and signatures ~typing_env ~reference ~current =
   let initialized_env = Typing_env.initialized_env () in
-  (*let modified_reference, modified_current =
+  let current =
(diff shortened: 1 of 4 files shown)


azzsal:
~~This also takes care of #84. We have to use fully qualified names, since that's how things are looked up in the environment.~~

